PR #15, opened 2015-11-11
Clarify summary challenge

cmacdonell:
When teaching this lesson I found the Summary challenge a bit unclear.  I've added a simple stacked bar plotting example to more clearly illustrate how dataframes are displayed when plotted in stacked bars.

Lastly, I also provided a solution to the Summary challenge.


01-starting-with-data.md
```diff
@@ -453,25 +453,100 @@ being sex. The plot should show total weight by sex for each plot. Some
...
+---------------------------
+
+
+The next two lines group data by plot and by sex, and then have sum the total for each plot.
 
-my_plot=data.plot(kind='bar',stacked=True,title="Total Weight by Plot and Sex")
-my_plot.set_xlabel("Plot")
-my_plot.set_ylabel("Weight")
+```
+by_species_sex = surveys_df.groupby(['plot','sex'])
```
hlapp: Shouldn't this be named `by_plot_sex`? It confused me for quite a while what the species is doing in here, but perhaps that name is only a remnant from an earlier version that needs to be fixed.

01-starting-with-data.md
```diff
@@ -453,25 +453,100 @@ being sex. The plot should show total weight by sex for each plot. Some
...
+4     F      39796
+      M      49377
+<other plots removed for brevity>
+```
+
+Below we'll use `.unstack()` on our grouped data to figure out the total weight that each sex contributed to each plot.
+
+```
+by_species_sex = surveys_df.groupby(['plot','sex'])
+sp_sx_count = by_species_sex['wgt'].sum()
```
hlapp: Again as above, should this be named `by_plot_sex`? And then `sp_sx_count` should be `plot_sx_count`, right? BTW I think there's no harm in retaining the vowel to make things super clear, i.e., why not use `plot_sex_count`?

01-starting-with-data.md
```diff
@@ -453,25 +453,100 @@ being sex. The plot should show total weight by sex for each plot. Some
...
+<other plots removed for brevity>
+```
+
+Now, create a stacked bar plot with that data where the weights for each sex are stacked by plot.
+
+Rather than display it as a table, we can plot the above data by stacking the values of each sex as follows
+
+```
+by_species_sex = surveys_df.groupby(['plot','sex'])
+pl_sx_count = by_species_sex['wgt'].sum()
```
hlapp: Again as above, should this be named `by_plot_sex`?

01-starting-with-data.md
cmacdonell: Yes, I'll fix these and push

cmacdonell:
Will fix, push and open again

PR #25, opened 2015-12-26
adding in link to the data for the lesson

tracykteal:
This addresses https://github.com/datacarpentry/python-ecology/issues/24

Added information on the dataset used and a link to the data in the 01-starting-with-data.md lesson and added data.zip so the whole directory can be downloaded.


tracykteal:
This needs to be updated now that the lesson has been updated to use the data on FigShare. PR on hold until this PR is updated.

PR #27, opened 2016-01-07
Use FigShare

### LINKED ISSUE #23: use the portal data from FigShare for this lesson
This lesson uses the portal data, but a copy that is in the repo, rather than the one on FigShare

https://figshare.com/articles/Portal_Project_Teaching_Database/1314459

It would be good to update the lesson to use this data. This will likely require some changes throughout the lesson though, since the FigShare version of surveys.csv has an added column 'hindfoot_length' and different column headers.

mkuzak:
I have adjusted lessons to use FigShare instead local files.
Some of used files are not on FigShare and therefore I left the way they are used in the code unchanged.
Fixes #23

### DIFF
--- a/01-starting-with-data.md
+++ b/01-starting-with-data.md
@@ -144,7 +144,7 @@ The above command yields the **output** below:
 [35549 rows x 8 columns]
 ```
 
-We can see that there were 33,549 rows parsed. Each row has 8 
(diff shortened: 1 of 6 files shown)


tracykteal:
Thanks, this was a change that really needed to be made, and I know a lot of changes were required. This is a great improvement.

PR #29, opened 2016-01-31
Use figshare fix

mkuzak:
In previous pull request I missed few things that does not work with figshare dataset. Mostly using `species` instead of `species_id`. So now I fixed those. Hope it's all of them.


ethanwhite:
Thanks @mkuzak!

PR #31, opened 2016-02-01
Remove "lists are the most common data structure" phrase

amueller:
I find the statement "lists are the most common data structure" without any qualifications a bit odd.
In particular, in data science, lists should not be used to store your data set. Operations on lists are waaayyy to slow. Data should be stored using pandas or numpy.
I'm not sure if that was meant. If you are thinking about general python, you could argue tuples are more common. However, that is not really relevant to the material here.

### DIFF
--- a/00-short-introduction-to-Python.md
+++ b/00-short-introduction-to-Python.md
@@ -111,7 +111,7 @@ True
 
 ### Lists
 
-**Lists** are the most common data structure. They can hold a sequence of
+**Lists** are a common data structure to hold a sequence of
 elements. Each element can be accessed by an index:
 
 ```python


hlapp:
Perhaps change that to _common_? Whether a list as a data structure is simple is in the eyes of the user - more advanced users will agree, novices (who we are targeting) likely won't.  Perhaps something similar to this: _**Lists** are often a convenient (even if not necessarily efficient) data structure for representing data._

amueller:
I'd be fine with "Lists are a common data structure" or "Lists are often a convenient data structure" - I agree with your point on simple. And I wouldn't add the "not efficient" bit in here, as it is too early in the course to discuss that, it was more a side node about why I wouldn't encourage the use of lists too strongly.

hlapp:
Sounds good. Do you want to go ahead and make that change to _common_ or _convenient_?

amueller:
done

PR #35, opened 2016-02-10
modified code description and column name

menegon:
This patch contains two simple fixes because.


tracykteal:
Thanks, it seems both these fixes improve accuracy and fix code that wasn't working. Is that right?

menegon:
yes, exaclty!

PR #38, opened 2016-02-29
Swcarpentry checkout

borsdorff:
challenge for matplotlib


apawlik:
@borsdorff Thank you for the PR though it looks like not the right repo and there seem to be lots of files which don't belong to the material. As per my earlier email, would you be able to correct that? Many thanks.

apawlik:
@borsdorff ouch sorry - the repo is correct. It's just the extra files that should not be there. Would you be able to inspect that?

PR #39, opened 2016-02-29
matplotlib challenge

borsdorff:
matplotlib challenge

### DIFF
--- a/06-plotting-with-matplotlib.md
+++ b/06-plotting-with-matplotlib.md
@@ -389,6 +389,14 @@ for group in grouped_plot_data:
 plt.legend()
 ```
 
+## Plotting a Parabola {.challenge}
+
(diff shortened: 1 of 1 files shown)


apawlik:
@borsdorff Thanks for cleaning it up! :-)

justbennet:
It might be a bit more colloquial English to use "...an array x with + 1000 equally spaced..."

Also, just a change to a bit more colloquial, maybe "Calculate the y=x**2 for each element of x, and plot y against x,  limiting the x axis range to -10 to 10."

Do those read well?

wrightaprilm:
Hi all-

Just cleaning up left over pull requests before the lessons get archived in Zenodo. We did a major rehaul of the plotting lesson, if you have any thoughts we'd love to hear them!

PR #42, opened 2016-03-10
Update 02-index-slice-subset.md

mboisson:
The original code does not "ask python to select rows that have a NaN value for weight". It "ask python to select the weight of rows that have a NaN value anywhere in them".


qjcg:
Thanks for this @mboisson ! The issue you raised has now been fixed in [07fdd9f](https://github.com/datacarpentry/python-ecology-lesson/commit/07fdd9fcf5433dc6e3097ebc4348a982a5181bca), with attribution to you in the commit message. Cheers!

Issue #24, opened 2015-12-26
add link to the data in the 'starting with data' lesson


tracykteal:
There is not currently a link to the data used in the lesson, in the '01-starting-with-data' lesson.  We should add a link to the data. 

That link is currently in the repo in https://github.com/datacarpentry/python-ecology/tree/gh-pages/data

When we switch over to the FigShare database as in https://github.com/datacarpentry/python-ecology/issues/23 we can update the link.

hlapp:
Changing the data link to Figshare still needs to happen, the merge of #25 (see 9ec7fc93e6bc4e2bf3769e91d41b283d50c055f2) didn't address that.

tracykteal:
It's now addressed in https://github.com/datacarpentry/python-ecology/pull/34 and should now be fixed

mboisson:
I am currently going through the lesson to teach it tomorrow. The links provided are completely out of phase with the content of the lessons themselves. For example, lesson on automating data refers to a "data" folder, which does not exist (only 2 files are provided for download, rather than a whole directory). It also supposedly contains the files "plots.csv", "portal_mammals.sqlite", "species.csv", "surveys.csv", etc. In previous lessons some of those files are used, but downloaded each time from figshare. 

It would be best if a single archive was downloaded, and if all the lessons would use the data that is contained within the archive.

ethanwhite:
@mboisson - I'm not actively involved in this lesson, but for the data itself is now centrally stored on figshare: https://figshare.com/articles/Portal_Project_Teaching_Database/1314459. If you think an archive with a particular set of files would be a useful addition please open an issues at https://github.com/weecology/portal-teachingdb. Thanks.

mboisson:
From what I can see, the following file structures is what is needed for the Python Ecology Lesson : [python-ecology]$ ls data/
portal_mammals.sqlite	species.csv		surveys.csv

The biggest part of the work however is to modify the lesson so that it uses those files rather than downloading the single files every time. It is more of a work on the Data Carpentry lesson than work on the organization of the files in Figshare.

justbennet:
Maxime, you get a couple of extra files, but the Download all button on the figshare site links to a .zip of everything.  Does that help any?

You said:  "The biggest part of the work however is to modify the lesson so that it uses those files rather than downloading the single files every time."  Do you mean, modify the lesson so the pd.read_csv() commands refer to the local files rather than to URLs?

mboisson:
[...]

PR #52, opened 2016-04-19
Modified 01starting-with-data.md to use .csv files instead of URLs

justbennet:
This is the cleaned up version of my instructor training pull request.  John slipped in a change to the
Python doc, which I incorporated, hence it appears there are two commits to this pull request instead of one.  This should result in all and only the changes I intended to make.  Sorry for the bumbling.

This addresses a concern expressed by @mboisson in Issue 24.

### DIFF
--- a/01-starting-with-data.md
+++ b/01-starting-with-data.md
@@ -12,7 +12,6 @@ title: Starting With Data
 * Load a Python/Pandas library.
 * Read tabular data from a file into Python using Pandas using `read_csv`.
 * Learn about the Pandas DataFrame object.
-* Learn about data slicing and indexing.
 * Perform mathematical operations on numeric data.
 * Create simple plots of data.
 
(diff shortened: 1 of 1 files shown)


qjcg:
Hi @justbennet : I notice that you still remove the "slicing and indexing" learning objective without replacing it with something more accurate. In fact, I think  updating that learning objective should probably be a separate PR.

justbennet:
OK.  I can restore it.  Is everything else good?

qjcg:
Yes, the rest looks good to me.

justbennet:
Should be good now.

qjcg:
Great, merged, thank you!

CC: @gvwilson

qjcg:
Also, quick note: though this PR is related to #24, it doesn't fully resolve it. I'll be teaching in early May and plan to look more closely at that issue before then.

justbennet:
@qjcg Yes.  I am hoping to get you another pull request with some suggested changes to the beginning of the lesson that should address the rest of the issue and, I hope, make the presentation a bit smoother.

qjcg:
Sounds good!

justbennet:
@qjcg, I submitted PR #54 just now, which is a revision of the first several sections.  I hope there is something there you can use.  I'll switch to that channel for further discussion. I hope I didn't overstep.

PR #53, opened 2016-04-20
Modified the learning objectives to be more complete

justbennet:
Removed the objective about slicing and indexing, as that moved to a different lesson.

Added some topics that are covered as objectives so that the coverage of the material by the objectives is more complete.

It's a bit disconcerting to have data about spatial plots from which plots are created.  I changed the wording to 'bar charts' in the objectives for better diferentiation.


wrightaprilm:
Hi @justbennet -

We're working on finishing up the PRs for the next archive of the lessons. Our new assessment specialist, Kari Jordan, did a full rewrite of the learning objectives. We'd love to hear your thoughts, if you're willing. They're now merged into each of the lessons on the gh-pages branch.

PR #55, opened 2016-05-07
Improve binary operator introduction examples

thomasballinger:
This adds another space before inline comments and changes the modulo example to use number that make it easier to deduce what "modulo" means from context.


qjcg:
Looks good to me, thanks @thomasballinger!

PR #57, opened 2016-05-07
add diagram with slicing index numbers

thomasballinger:
What do you think of this diagram, and does it fit here? I'm happy to make any changes or scrap it. I think the exercises after this point still make sense with the diagram although I suppose the presence of a diagram means folks might not realize they need to be trying these examples live.

re: diagram specifics, I think it's a bit noisy; I'd be happy to pull out the code examples and/or the "slicing: ..." texts.

image since it doesn't seem to show up for me in the diff preview on GitHub:


thomasballinger:
Hm regarding [the svg not showing up on GitHub .md files](https://github.com/isaacs/github/issues/316), should I link to [something like I did above](https://rawgit.com/) so the page still looks good on GitHub? Or is it ok to look bad for now, but then later link to the github.io url?

I was going to try this but found the link down, https://github.com/datacarpentry/python-ecology-lesson/issues/58

wrightaprilm:
I find this a little noisy. What if we separated this into two images: one for indexing, and highlighted 92 on red, and one for slicing with the elements of the slice highlighted? Then the subscripts could be right under the element without worrying about that being confusing.

thomasballinger:
Here's also a version with no text, so not focused on teaching vocabulary but just elements vs space between elements for slicing:
simpler

I like that more visual thing you describe. I'll have to play with it a bit. To clarify, it sounds like you're describing a code sample like `grades[1:3]` accompanying a blue-highlighted slice, and a similar one for indexing? Would you prefer to keep the images separate from the code samples then?

wrightaprilm:
I like the code samples on the image (but open to alternatives). I just think separating the two examples will make it easier to draw the eye to the differences between the effect of the code samples.

thomasballinger:
I've separated indexing and slicing into two diagrams. I suppose they still belong together here, because although indexing and slicing are addressed separately in the lesson this is the place where we use simple Python lists to demonstrate 0-indexing.

indexing

slicing

Happy to make more changes, apologies for the delay as I was out hiking for a week!

thomasballinger:
Rereading @wrightaprilm's feedback I see I missed the bit about only using one set of subscripts. I like having both to highlight that there's a difference, but can certainly change this. I also forgot about the color suggestion of red, happy to swap out the colors for whatever makes sense.

wrightaprilm:
Oh, I meant one index per image. I don't actually know what the two indices are supposed to represent (inclusive/exclusive?)?
This looks much better - I think swapping the black border on the colored box for transparent and bolding the numbers, rather than circling them will help with the visual overload.

thomasballinger:
Hm, yeah I meant for the second set of indices to show how "inclusive at the bottom, exclusive at the top" behavior works without using those words, instead thinking of the indices as standing for the gaps *between* the elements.

Current versions:





From here, which of these should I do:

- [ ] delete the unused set of indices for each
  - [ ] and move the other set of indices to the
    - [ ] top
    - [ ] bottom

thomasballinger:
(thanks for your patience by the way with this feedback, and my apologies for my long response times despite your short ones!)

wrightaprilm:
Voted 👍 

No problem, btw. I do most of my maintainer stuff on Saturdays, since I don't like to do it during the workweek, and my daughter has an activity on Saturday AM that gives me some free time. I think our schedules just happen to be synced up well right now.

thomasballinger:
Checking in on this, let me know if there's anything else I should change. Thanks!

wrightaprilm:
Sorry, missed your last message. This looks good to me. How about you, @qjcg?

qjcg:
Looks good to me too, and I think it's great to have something visual for this. Thanks @thomasballinger!

thomasballinger:
Thanks both, happy to share a favorite diagram.

PR #62, opened 2016-05-25
Make a loop over values that must be pulled from the data. - Instructor training assignment

sdtaylor:
The student has already used `pd.unique()` to see all unique species, so they could copy and paste that into a list here to iterate over. But hopefully they will realize that they can make a list directly from `pd.unique()` to use in the for loop. A good concept of letting the code discover what needs to be done. 
Ideal answer:

```all_species_ids=surveys_df['species_id'].unique()
for species_id in all_species_ids:
    survey_spp=surveys_df[surveys_df.species_id == species_id]
    filename = 'data/surveys' + str(species_id) + '.csv'
    survey_spp.to_csv(filename)
```

### DIFF
--- a/05-loops-and-functions.md
+++ b/05-loops-and-functions.md
@@ -279,7 +279,7 @@ imagine we had used 1976 as the start year in `range`)?
 
 3. Let's say you only want to look at data from a given multiple of years. How would you modify your loop in order to generate a data file for only every 5th year, starting from 1977?
 
-4. Instead of splitting out the data by years, a collegue wants to do analyse each species seperatly. How would you write a unique csv file for each species?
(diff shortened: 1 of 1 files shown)


wrightaprilm:
I think in this iteration of the lessons, species_id is now just species. When I change surveys_df.species_id to surveys_df.species, I get the desired result. Is that right, or have I misunderstood something?

I do like this clarification, and another instructor had suggested similar, but I had not made the changes yet.

sdtaylor:
All the other lessons in the commit I forked have it as species_id. And the raw data link in lesson 1 has the column as species_id.

wrightaprilm:
I'm sorry, @sdtaylor, I thought I had merged this last month. Thanks for your contribution!

PR #64, opened 2016-05-25
Revise wording and example in "for" loop introduction

ceholden:
PR makes two alterations to the `for` loop lesson:

1. Wording choice nitpick
    * "returns" is used as in "carriage return", but "return" is also used in this lesson to refer to the `return` keyword. I'm hoping "new line" might avoid any potential for learners to be tripped up by the challenge.
2. The first introduction example to a `for` loop introduces two concepts at once -- a `for` loop and the idea that `str` is a collection that can be iterated over
    * Proposed change substitutes example datatype with a `list`, something that had been already explained as a collection in the `00-short-intro...`
    * Not included, but I wonder if we might introduce the `len("a string")` and `"string"[0]` concepts in the introduction lesson
    * On a related note, `00-short-intro...` shows examples of `for` loops over `list` and `dict` types before this loop type is introduced in `05-loops-and-functions.md`. 

I'd be happy to address the last two bullet points I wrote in item 2 if it would be useful.

PR submitted as part of instructor training from session with Greg Wilson on 4/14/2016.

### DIFF
--- a/05-loops-and-functions.md
+++ b/05-loops-and-functions.md
@@ -69,7 +69,7 @@ the loop. The statement `pass` in the body of the loop just means "do nothing".
 
 1. What happens if we don't include the `pass` statement?
 
-2. Rewrite the loop so that the letters are separated by dashes, not returns
+2. Rewrite the loop so that the letters are separated by dashes, not new lines
 (Hint: You can concatenate strings using a plus sign. For example,
 `print(string1 + string2)` outputs 'string1string2').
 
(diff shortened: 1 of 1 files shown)


qjcg:
Hi @ceholden, and thanks for your contribution!

This looks good to me. I've seen students struggle with the for loop and string iteration as combined together here, and you're right that this adds unnecessary complexity when introducing the for loop.

Regarding the last two points you raised but have not addressed here: please submit a PR and we'll happily review it!

PR #63, opened 2016-05-25
revised learning objectives for data types and formats lesson

valpasq:
Created a new set of learning objectives to better match lesson content. Grouped and labeled new objectives by level in Bloom's taxonomy.

PR #66, opened 2016-05-27
group learning objectives into two main levels

valpasq:
PR groups learning objectives into two (rather than five) levels, as per
comments on initial PR (https://github.com/datacarpentry/python-ecology-lesson/pull/63) to update learning objectives. This two-level format is encouraged going forward to reduce visual clutter for learners.


qjcg:
We're off to a great start --- thanks @valpasq!

PR #68, opened 2016-06-11
Make instructions a little clearer

Deborah-Digges:
The original text is "Let's next create a list of unique species in our data", which would need a statement like
` df.species_id.nunique()`, when in fact what we are doing is `surveys_df.groupby('species_id')['record_id'].count()`, which is the number of samples in each species.


qjcg:
LGTM. Thanks @Deborah-Digges!

PR #69, opened 2016-06-22
added challenge question on dictionaries to 00

SeanRGB:
This is for DC Instructor Training checkout. @gvwilson


wrightaprilm:
This looks good to me. How about you, @qjcg?

qjcg:
Seconded and merged, @wrightaprilm. Thanks for contributing @SeanRGB!

PR #70, opened 2016-06-30
Added a 3rd exercise to lesson 05, loops challenge.

gabayae:


gabayae:
This third example is meant to use the "for loop" on a different type of container, very useful in practice, namely "dictionaries".

wrightaprilm:
I'm a little hesitant on merging this one. I think you've done a good job with the content, but dictionaries are emphasized so little in the Data Carpentry materials that I'm worried that bringing them back five lessons after their introduction might be a little confusing. I'm trying to think what I think would be useful here, and I've come up with two options:

1. We could put in parentheses a short reminder of what dictionaries are and link back to lesson00. 
2. In the text, we could remind users of what they are and why we use them.

If we're going to merge this, I favor solution 1, since I have a feeling that we might be getting a little overstuffed. Thoughts, @qjcg? Was this lesson over full when you ran it last month?

gabayae:
Hi **@wrightaprilm**., thank you for your comment and suggestions.

I realized, later on, that Data Carpentry materials do not emphasize on dictionaries, but have at least introduced them. On that note, in my opinion, solution 1, seems indeed appropriate.

PR #76, opened 2016-08-08
added information about indexes in a dataframe

snamburi3:
The DataFrame index was not explained in the lesson. I have added some information about that.


wrightaprilm:
Looking good, thanks for your contribution! Would you mind adding a little bit to the sentence where you say that the index is not actually a column in the dataframe explaining what the index _is_? I think _is not_ statements work best when paired with an _is_. 

Thanks!

snamburi3:
You are right! In the new commit, I explained about what an index is.  what do you think of the new sentence?

wrightaprilm:
I like that a lot! Look good to you, @qjcg?

Thanks for contributing! Is this for instructor training?

qjcg:
Looks good to me! Thanks for the ping @wrightaprilm, and thanks for contributing @snamburi3!

PR #77, opened 2016-08-11
Add function exercise to episode 01

aschoenr:
Added function exercise to episode 00. This exercise asks the student to write a function that takes two lists as parameters as return a dictionary with the first list as the keys and the second list as the values. This will force the student to use many of the ideas presented in a relatively short function. This submission was done as a part of the instructor training checkout steps.

### DIFF
--- a/00-short-introduction-to-Python.md
+++ b/00-short-introduction-to-Python.md
@@ -182,7 +182,6 @@ another_tuple = ('blue','green','red')
 # notes lists uses square brackets
 a_list = [1,2,3]
 ```
-
 ### Challenge
 1. What happens when you type `a_tuple[2]=5` vs `a_list[1]=5` ?
 2. Type `type(a_tuple)` into python - what is the object type?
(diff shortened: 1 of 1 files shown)


wrightaprilm:
This looks good to me. I am a little worried about the length of this lesson, though. When you taught it, how did it go, @qjcg? Should we pull something out to merge this in? Since this is a combined function-dictionary question, I would be OK replacing our current dictionary question with this one. 

I've opened issue #81 to discuss if the lessons are overfull or not. I'm going to hold off on merging this unless anyone else feels strongly.

willingc:
@wrightaprilm I actually like this suggested Challenge by @aschoenr as an optional take-home or advanced challenge.

wrightaprilm:
Since no one else has weighed in on lesson 1 being overfull, I'm going to add this in to the lesson. I would normally handle this by dictionary = dict(zip(keys, values)). But perhaps for this audience, I should break this out into a loop in the instructors guide? Or point them to use the ? for help with the dict and zip functions? Thoughts?

tracykteal:
Thanks all. Please hold off on merging until after the lesson release this week.

Issue #44, opened 2016-03-21
Why not use surveys_df['year'].unique() rather than range()


mboisson:
In the lesson about automation, why not pull the list of unique years from surveys_df['year'].unique() rather than creating confusion with the range() function ?

It would simplify the lesson quite a bit.

itcarroll:
I believe @qjcg meant to close this issue with pull request #78.

tracykteal:
It has been addressed, so I'll close this.

PR #78, opened 2016-08-17
Addresses issue #44

cgeroux:
I addressed Issue #44 . As suggested this simplifies the instruction and also guards against the possibility of missing years. However, if the point was to introduce the range function then perhaps this change should not be incorporated.


wrightaprilm:
I don't actually feel strongly about introducing the range function, so I quite like this. If you're good with it, @qjcg, I think we should merge this.

qjcg:
Looks good to me @wrightaprilm! Thanks for your contribution @cgeroux, merged.

Issue #73, opened 2016-07-26
Challenges are not always clear


ostueker:
I'm just working myself through this lesson, as I haven't worked with pandas so far and I found that for most of the challenges, it is not clear to me which result is expected.

Especially for challenges that ask to generate a plot, it would be really helpful if a PNG of the desired plot would be added to the challenge.  

I wanted to help adding solutions to the challenges to the `INSTRUCTORS.md` but often it is not clear to me what the person who has designed the challenge had in mind.

wrightaprilm:
I agree - the creation of an Instructor's Guide fell by the wayside with a lot of the Data Carpentry lessons. Since, for some of the challenges, it can be unclear what the desired output is, I think the best way forward is to make a pull requests to the guide, individually for each solution/answer. That way, we can take a look, and if there are multiple possible solutions, I (or someone else) can add those in and do the merge. 

Thanks for bringing this to our attention - and sorry for my late reply. Summer travel season strikes again!

tracykteal:
The challenge solutions are now in the instructors guide. Hooray! Is there more to do here though?

wrightaprilm:
I believe #171 and #162 close these. Thanks to @tmorrell

PR #82, opened 2016-09-06
Provide Instruction notes to solutions of the challenges

stijnvanhoey:
With respect to issue #73, this patch provides a first draft of (possible) solutions to the challenges from lesson 01 till 05. It includes the resulting graphs (img/ foldre) in default matplotlib styling and the code to produce these plots.


wrightaprilm:
These look really great, thank you so much for tackling these. I'm going to go ahead an merge this PR.

Thanks so much for your contribution!

PR #86, opened 2016-09-22
Scaffold for Instructor's guide and solutions for the first exercises.

ostueker:
As I've said in issue #73, the challenges are not always clear and so far no example solutions exist.

This PR contains a scaffold for the Instructor's guide an example solutions for the a few (still straight forward) exercises (module: Starting with Data).

### DIFF
--- a/INSTRUCTORS.md
+++ b/INSTRUCTORS.md
@@ -1,0 +1,81 @@
+---
+layout: page
+root: .
+title: Python for Ecologists
+subtitle: Instructor's Guide
+---
+
+
+
+## [Short Introduction to Programming in Python](00-short-introduction-to-Python.md)
+
+## [Starting With Data](01-starting-with-data.md)
+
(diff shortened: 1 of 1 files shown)


wrightaprilm:
Hi @ostueker -

It looks like there are some conflicts between the instuctor's guide we have now, and this one. So I'll get to the merge on this mid-to-late-week next week, unless you need for it to be handled before then. Does that sound OK?

ostueker:
A PR that I've overlooked got merged short after my last pull. That one includes more solutions than I've been able to create this morning.

Let's cancel this PR and I look which of my additions still make sense to merge.

wrightaprilm:
OK, cool. I might do a little restructuring of where we're stashing the sample outputs for the exercise where the learners break up a spreadsheet by year, but nothing should affect you working on the guide itself. Thanks for looking at this!

Issue #81, opened 2016-08-31
Lessons 00/01 getting a little overfull


wrightaprilm:
This round of instructor training, we had a lot of interest in the first two lessons in the Python set. Which is great! But also puts us in a position where we're looking a little overfull. So the first thing I want to know is if that's true. I think @qjcg has taught through this lesson recently? How did you find 00 and 01 went? Should we have a further discussion about streamlining these, or are we doing OK?

tomhohenstein:
Hello!

Are you still looking into trimming lessons 00/01? I'm going through the  instructor training checklist and am willing to help. I also noticed that lesson 00 does not have learning objectives (I'd be happy to add that)

wrightaprilm:
We are still interested in doing that, and would greatly appreciate the help. If you're interested in the issue to learning objectives, we have a separate branch for re-organizing them into Bloom's Taxonomy (suggested by @ErinBecker in #65). That's been a little stalled, for various reasons, but you're absolutely welcome to have a crack at it.

wrightaprilm:
Since #169 is merged, I'm going to close this.

PR #94, opened 2016-11-30
01 lesson trim

tomhohenstein:
Hello! 

I'm starting a pull request to trim down lesson 01 mentioned in #81. I'm starting this pull request now so that others can offer input into the process. 

My first goal is to consolidate material. Then, based on others input, we can think about how much plotting we should do. 

Any advice or suggestions are greatly appreciated.

### DIFF
--- a/01-starting-with-data.md
+++ b/01-starting-with-data.md
@@ -75,7 +75,6 @@ record_id,month,day,year,plot_id,species_id,sex,hindfoot_length,weight
 
 ---
 
-
 ## About Libraries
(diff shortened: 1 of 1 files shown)


wrightaprilm:
Sorry I'm so late getting back to this. These commits look great, I'll merge them right now. Thanks so much. Also, is this instructor training? If so, I'll go ahead and tag it as such.

tomhohenstein:
It was instructor training - thank you!

Issue #99, opened 2016-12-09
No images of plots in 06-plotting-with-matplotlib


jarmokivekas:
I noticed that lesson [06-plotting-with-matplotlib](http://www.datacarpentry.org/python-ecology-lesson/06-plotting-with-matplotlib) does not include any images of the plots that would be created by the various bits of example code. Is this a design decision, or have the images just not been created? I'd be willing to create python script that created example plot, and link those into the lesson's markdown.

wrightaprilm:
Hi @jarmokivekas -

Closing this, as we've revamped the lesson. Thanks for participating in the discussion with us!

PR #101, opened 2016-12-12
lesson 06: matplotlib, add plot images

jarmokivekas:
I commented about this in issue #99 a few days ago. I had some spare time and wanted to test out the new 2.0rc1 version of matplotlib, so I made a script that draws the plots based on the lesson's example code and saves them in `/img` as png files.

I don't know of a way to publish github-pages from a branch, but i published the `_site` directory here:
- <http://guttula.com/python-ecology-lesson/06-plotting-with-matplotlib.html>

---

I'm almost certain that the `plt.boxplot() ` example is wrong. The example code produces the plot below, which, is not really anything.

---

I aslo amended the tick positioning and re-labeling example to make sense with the dataset used in the examples.

Any thoughts? I can re-render the plots with the old matplotlib defaults if needed, since these have the new matplolib 2.0 default style.


06-plotting-with-matplotlib.md
```diff
@@ -341,6 +364,18 @@ To use a bar plot:
 plt.bar(plot_data.index, plot_data.values)
 ```
 
+
+![](img/06_plot_variations_bar_plot.png)
+
+A box and whisker plot:
```
jarmokivekas: This boxplot example was moved from the "Go exploring" section here. It is probably also wrong.

06-plotting-with-matplotlib.md
```diff
@@ -341,6 +364,18 @@ To use a bar plot:
...
+
+```python
+plt.boxplot(plot_data.values)
+```
+
```
jarmokivekas: This boxplot example was moved from the "Go exploring" section here. It is probably also wrong.

wrightaprilm:
Hi Jarmo,

Could I have you contribute to the discussion on #96? Plotting has always been a weakness of this lesson, and when I came on as a maintainer ...  well, I can make plots, but it's not a particular strength of mine. So I'm soliciting feedback on how/if we want to do a more serious rewrite of the plotting materials.

Thanks for your contribution; I think we can make something really cool with this part of the lesson!

PR #102, opened 2016-12-14
fixing naming conflict

aykol:
Since "sorted" conflicts with the built-in python function sorted(), renamed the example variable name as "sorted_data"


wrightaprilm:
Good catch. Thanks for contributing!

PR #109, opened 2016-12-16
Revised learning objectives

kariljordan:
@qjcg  @wrightaprilm  @mkuzak 
We are in the process of developing our surveys so that we can have a better understanding of our learners' skills and self-efficacy after taking a Data Carpentry workshop. To do this we want to improve the learning objectives for all of our lessons so that they are measurable. I revised the learning objectives with Erin's help for the Python lesson, and Kate advised that the best way to implement these types of changes would be to submit them as a pull request. I would love to have your feedback and input as maintainers of this lesson, especially with respect to whether we captured the terminology in the LOs correctly. Are you able to take a look at the learningobjectives/PRs and provide feedback by December 28th?​

### DIFF
--- a/06-plotting-with-matplotlib.md
+++ b/06-plotting-with-matplotlib.md
@@ -4,6 +4,12 @@ root: .
 title: Plotting Your Data - Matplotlib
 ---
 
+## Learning Objectives
+
+* Describe what the Matplotlib library is used for.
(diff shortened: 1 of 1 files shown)


wrightaprilm:
Closed, per discussion in #103 .

PR #74, opened 2016-07-28
renamed variables from CamelCase to snake_case

ostueker:
As suggested in issue #71, variables in segments 0-4 have been
renamed from CamelCase to snake_case to standardize the coding style.

PR #131, opened 2017-01-03
Few fixes improving style consistency.

mkuzak:
Few variable names missed in pull request #74 
Added point to CONTRIBUTING.md about adhering to PEP8


wrightaprilm:
Thanks for these. A number of us who worked on drafts of the materials do a bit of C/C++ and we get some camelCaseCreep that can be hard to root out.

PR #133, opened 2017-01-08
Replace 'batteries-included' with 'general-purpose'

### LINKED ISSUE #30: "Batteries included" is a lie for data science
The intro says

> "Batteries Included" philosophy - libraries for common tasks available in standard installation

I would argue that the opposite is true. Coming from R or Matlab, people are very surprised that you need to install / import a library (numpy) to get matrices. And a different one for dataframes.
Installing numpy also is a major pain on most operating systems, unless you use anaconda.
Does "standard installation" mean anaconda? Then the instruction should say that.

A novice might thing "standard installation" means the one from python.org. Imagine someone on windows installing the python.org interpreter and expecting batteries included.

[for context: I'm a strong advocate for the use of Python in data science.]

mkuzak:
fixes #30

### DIFF
--- a/00-short-introduction-to-Python.md
+++ b/00-short-introduction-to-Python.md
@@ -19,8 +19,7 @@ Python's main advantages:
 
 * Open Source software, supported by Python Software Foundation
 * Available on all platforms
-* "Batteries Included" philosophy - libraries for common tasks available in
-  standard installation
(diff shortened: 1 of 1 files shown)


qjcg:
Merged, thanks Mateusz.

PR #169, opened 2017-04-06
Edit Python overview

### LINKED ISSUE #143: Expand overview of functions in 01-short-introduction-to-Python
Currently built-in functions are used without introduction, then the lesson jumps to user-defined functions. 

I'm aware that capacity within lessons is limited, but I wonder if it would be helpful to alter the final section (## Functions) so that it minimally explains i)  the key concepts of a function, ii)  that we have already been using functions (print, dir, help) that are built-ins, iii)  and subsequently that it is possible for a user to write their own functions. 

P.S. I've just noticed that this issue has already been raised (#77), but was it ever resolved?

willingc:
Closes #143.

Addresses #81.

- Correct Python errors and inaccuracies.
- Update style to be consistent with Python docs and PEP8
- Add a tip addressing functions for #143

### DIFF
--- a/_episodes/00-short-introduction-to-Python.md
+++ b/_episodes/00-short-introduction-to-Python.md
@@ -92,6 +92,10 @@ $ python example.py
 Data Carpentry
 ```
 
+**Tip**: The `print` function is a built-in function in Python. Later in this
+lesson, we will introduce methods and user-defined functions. The Python
+documentation is excellent for reference on the differences between them.
+
 ### Operators
 
 We can perform mathematical calculations in Python using the basic operators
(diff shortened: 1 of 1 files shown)


tracykteal:
Thanks, this is terrific! Besides having this in Setup, I think it would be good to have some of that where instructors knew to go through it during the workshop.

Maybe we can leave that PR as is, and either some of it can be added to 'intro to Python' at the very beginning, or it could be its own episode, and it could be the first one taught, so 00-intro-to-jupyter-notebook? (I'd just have to figure out what happens if you have more than one 00 lesson on ordering.)

willingc:
Let's merge and then do a separate issue that reorganizes the way you would like it.

wrightaprilm:
I would prefer to merge this as is and do reorganization later, if need be. I'm catching up on PRs now - are all the lessons getting 00-level intros to the tools used (Jupyter notebook, rstudio, whatever SQL does)?

wrightaprilm:
Reads well; great work. Merging now.

willingc:
Thanks @wrightaprilm. Happy to help.

Issue #161, opened 2017-03-17
02 challenge questions


tmorrell:
These challenge questions

Challenge Activities
What happens when you type:
surveys_df[0:3]
surveys_df[:5]
surveys_df[-1:]
What happens when you call:
dat.iloc[0:4, 1:4]
dat.loc[0:4, 1:4]
How are the two commands different?

Are the exact same as the examples provided in the lesson

willingc:
I'm working on this issue along with #160

Issue #147, opened 2017-03-16
Check variables


wrightaprilm:
Thank you for filing an issue! 

These lessons are being prepared for publication. From March 15, 2017 to April 21, 2017 issues related to [lesson release preparations](https://github.com/datacarpentry/lesson-release/blob/master/release-checklist.md) are particularly appreciated.

During Bug BBQ, we could really use for someone to walk through the materials and make sure variables are named consistently. 

**Level:** Beginner.

wrightaprilm:
I think I have caught all the name inconsistencies.

PR #173, opened 2017-04-06
Lesson 2 - Revise the copy and reference section

### LINKED ISSUE #160: 02 copy example
Adjust text to suggest that the first copy attempt won't work

We can also reassign values within subsets of our DataFrame. But before we do that, let’s _try to _ make a copy of our DataFrame so as not to modify our original imported data.

```
#try to copy the surveys dataframe so we don't modify the original DataFrame
surveys_copy = surveys_df
```

```
#set the first three rows of data in the DataFrame to 0
surveys_copy[0:3] = 0
```

_At the end of the next section, repeat example code that shows a successful copy_

```
#set the first three rows of data in the DataFrame to 0
surveys_copy[0:3] = 0
surveys_copy.head()
surveys_df.head()
```

willingc:
Closes #160 and #161 

This is still a WIP. Copy and references are one of the most confusing concepts for those new to them. I believe it is better to be direct and precise in explaining these two concepts than to have the students experiment to figure out the difference.

Feedback welcome on this draft.

### DIFF
--- a/_episodes/02-index-slice-subset.md
+++ b/_episodes/02-index-slice-subset.md
@@ -308,17 +308,21 @@ selects the element that is 3 rows down and 7 columns over in the DataFrame.
 
 1. What happens when you execute:
 
-- `surveys_df[0:3]`
-- `surveys_df[:5]`
-- `surveys_df[-1:]`
+    - `surveys_df[0:1]`
+    - `surveys_df[:4]`
+    - `surveys_df[:-1]`
 
 2. What happens when you call:
 
     - `dat.iloc[0:4, 1:4]`
     - `dat.loc[0:4, 1:4]`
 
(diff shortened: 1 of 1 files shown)


tmorrell:
I think this is a great set of revisions.  The copy vs reference section is much clearer.

willingc:
I've editing the entire lesson and am removing the WIP designation. Please review. Thanks!

tracykteal:
Agree, the language and formatting adds a lot of clarity.

wrightaprilm:
OK, I think we're good to go. I'm working through #147 right now, so I'm going to merge these ahead so I can read these changes as I do that.

Thanks so much!

willingc:
Thanks @wrightaprilm @tracykteal @tmorrell for the review and merge.

PR #175, opened 2017-04-07
Highlight in bold the first mention of each method

### LINKED ISSUE #87: starting-with-data: 2nd plotting example uses constructs that were not introduced
I have several issues with these two lines of code in the **Quick & Easy Plotting ...** section in the **Staring with data** module:

``` python
total_count=surveys_df.record_id.groupby(surveys_df['plot_id']).nunique()
# let's plot that too
total_count.plot(kind='bar');
```
1. So far it the `df.groupby()` method has only been given a single column name or a list of them. Suddenly and without explanation a whole column (`surveys_df['plot_id']`) is passed as a parameter.
2. The `.nunique()` method has not been introduced.
3. So far only the dict like notation for accessing columns in DataFrames has been used, now it's the attribute notation (actually in **Calculating Statistics ...** it is also used but not explained: `pd.unique(surveys_df.species_id)`)
4. (style) no whitespace around `=`

I'm still new to pandas and am right now at my second round through this module (on my own) and this is one of the points that easily throws me off. 

My suggestions:
- `.nunique()` could be quickly mentioned right after `.unique()`
- the dict-notation should be used consistently (until the attribute notation is introduced at a later point).

(I'll hand in a PR soon for those.)

I'm not sure how to deal with the `.groupby()` though. I'll have to read up myself on this.

willingc:
Closes #87 

This closes the issue re: `groupby` being used before mentioning. I've gone ahead and bolded the first mention of all the methods to make it easier for folks to scan back through the lesson.

### DIFF
--- a/_episodes/01-starting-with-data.md
+++ b/_episodes/01-starting-with-data.md
@@ -34,7 +34,7 @@ A quick aside that there are Python libraries like [OS
 Library](https://docs.python.org/3/library/os.html) that can work with our
 directory structure, however, that is not our focus today.
 
-### Our Data 
+### Our Data
 
 For this lesson, we will be using the Portal Teaching data, a subset of the data
 from Ernst et al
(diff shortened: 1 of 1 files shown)


wrightaprilm:
Merging this now, thank you!

PR #176, opened 2017-04-07
Challenge formatting

### LINKED ISSUE #159: 02 lesson cleanup
Change order of questions to make more sense

Challenges
What value does the code below return? a[0]
How about this: a[5]
**In the example above, calling a[5] returns an error. Why is that?
What about? a[len(a)]**

Remove repeated command (italics)

_select the first, second and third rows from the surveys variable
surveys_df[0:3]_
select the first 5 rows (rows 0,1,2,3,4)
surveys_df[:5]
select the last element in the list
surveys_df[-1:]

tmorrell:
Fix challenge question formatting for lesson 02 and resolve #159


willingc:
Looks good to me @tmorell. Ready to merge @wrightaprilm @tracykteal

wrightaprilm:
Just ran this all, looks good. Thanks!

PR #180, opened 2017-04-22
ggplot

tmorrell:
Include install instructions for ggplot and improve formatting

### DIFF
--- a/_episodes/06-visualization-ggplot-python.md
+++ b/_episodes/06-visualization-ggplot-python.md
@@ -71,7 +71,11 @@ To build a ggplot we need to:
 - define aesthetics (`aes`), by selecting the variables to be plotted and the variables to define the presentation
      such as plotting size, shape color, etc.,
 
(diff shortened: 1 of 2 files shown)


wrightaprilm:
Hi Thomas, having a quick look at this. Mostly looks good, but I was wondering if bokeh is required by ggplot? I was under the impression that it is not. If that's not the case, we can have the learners install it, but if it's not required, let's not include it. 

Also, about removing the inline statement. I'm fine with that, how would you feel about including %matplotlib notebook, which allows for some handy plot utilities, like zooming on plots or exporting though point-and-click?

tmorrell:
At least on my current anaconda install ggplot is not present in the main channel (conda install ggplot doesn't work). The -c on the install line refers to the bokeh channel and will only install ggplot.  There might be a better option though?

Thanks for the %matplotlib notebook suggestion-I've added it to the PR.

wrightaprilm:
Aha, I see the issue. Well, this looks good to me; thank you! Merging now.

Issue #138, opened 2017-01-16
Add timing estimates to each episode


qjcg:

tracykteal:
Also related to the schedule in https://github.com/datacarpentry/python-ecology-lesson/issues/91

Adding timing will create a schedule.

tmorrell:
I'm teaching this lesson next week and will provide my timing as a starting point.

tracykteal:
Great, thank you! That will be really helpful.

tracykteal:
Any other feedback would be welcome too. We won't change things too much before release next week, but if things could be reordered, expanded or reduced, we can take that on after the release for the next version.

tmorrell:
Will do!

tracykteal:
I added initial estimates reading through the lessons.

https://github.com/datacarpentry/python-ecology-lesson/pull/183

 It will be good to know how that matches with teaching time. Also, from just these estimates, it looks like there's too much to teach in one day (a morning session + an afternoon session). I'll create another issue where we can discuss if there's too much to teach, and if so, what are people leaving out.

qjcg:
Closed by #183.

PR #183, opened 2017-04-25
added time estimates

tracykteal:
Added initial timing estimates and the number of challenges for the episodes

A first pass at https://github.com/datacarpentry/python-ecology-lesson/issues/138

### DIFF
--- a/_episodes/01-starting-with-data.md
+++ b/_episodes/01-starting-with-data.md
@@ -1,7 +1,7 @@
 ---
 title: Starting With Data
-teaching: 0
-exercises: 0
+teaching: 60
+exercises: 6
 questions:
     - " How can I import data in Python?"
     - " What is Pandas?"
--- a/_episodes/01-starting-with-data.md
+++ b/_episodes/01-starting-with-data.md
@@ -1,7 +1,7 @@
 ---
 title: Starting With Data
-teaching: 60
-exercises: 6
+teaching: 30
+exercises: 30
 questions:
     - " How can I import data in Python?"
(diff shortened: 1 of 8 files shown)


qjcg:
Thanks @tracykteal for going through and doing these initial estimates!

Overall LGTM, but I suspect the timing for the exercises is underestimated.

I suggest we allocate exercise timings in units of either 5 or 10 minutes to account for instructor explanations, the inevitable learner questions and issues that arise. Allocations of 2-3 minutes for an exercise in my experience would rarely be realistic.

@wrightaprilm (and anyone else reading), what are your thoughts?

tracykteal:
Ha! I was doing it as number of exercises, rather than amount of time. Were you wondering why I thought things were taking 6 minutes? :) I'll go update that splitting the time with teaching and challenges.

tracykteal:
I updated the PR. In general it seems there's slightly more than 50% of any lesson for challenge time, so I estimated things that way.

qjcg:
Haha... well knowing now that before you were listing the number of exercises and not the timing in minutes, that makes *way* more sense.

The updated timings look great!

I don't know if there's an "official" way we can ask instructors to comment / suggest updates to our timing estimates during or after a workshop (ex: in an email from us?), but I think it would be a good idea for keeping these timings realistic going forward.

Merging these initial estimages for now. Thanks!

tracykteal:
Asking for time estimate corrections could be something we do in discussion sessions. I'll suggest that as an idea to the mentoring group.

Issue #164, opened 2017-03-31
[Setup] ggplot fails to install from conda-forge using python 3.6 / Anaconda 4.3.1


omsai:
It looks like ggplot is only compatible with python 3.4 and not 3.6?

On March 30, 2017, a user reported installing the Python ggplot package fails to install using conda-forge.  Below was their report:

``` sh
MWC02N71WMG3QT:~ vn54$ conda install -c conda-forge ggplot
Fetching package metadata .............
Solving package specifications: .
UnsatisfiableError: The following specifications were found to be in conflict:
  - ggplot -> python 3.4*
  - python 3.6*
Use "conda info <package>" to see the dependencies for each package.
MWC02N71WMG3QT:~ vn54$ conda info <package>
-bash: syntax error near unexpected token `newline'
MWC02N71WMG3QT:~ vn54$ 
```

Sandeep Namburi confirmed this is the case for both macOS and GNU/Linux and suspects it may also impact Windows.

As a workaround for getting attendees setup for the workshop, in some cases one use `pip` instead:

``` sh
pip install -U ggplot
```

`pip` is provided by anaconda, but the user would need to have chosen add to anaconda to their `PATH`.

However in other cases, it `ggplot` complains of missing PyQt4.

I don't use anaconda and use my distro + pip packages instead, so I played with it in a VM:

``` sh
user@ubuntu16:~$ wget https://repo.continuum.io/archive/Anaconda3-4.3.1-Linux-x86_64.sh
user@ubuntu16:~$ bash Anaconda3-4.3.1-Linux-x86_64.sh
# Installed using the defaults options
user@ubuntu16:~$ export PATH=/home/user/anaconda3/bin:$PATH
user@ubuntu16:~$ conda install -c conda-forge ggplot
Fetching package metadata ...........
Solving package specifications: .
UnsatisfiableError: The following specifications were found to be in conflict:
  - ggplot -> python 3.4*
  - python 3.6*
Use "conda info <package>" to see the dependencies for each package.
user@ubuntu16:~$ conda info ggplot
Fetching package metadata .........
NoPackagesFoundError: Package missing in current linux-64 channels: 
  - ggplot
# One can see package info from https://anaconda.org/conda-forge/ggplot
user@ubuntu16:~$ which pip
/home/user/anaconda3/bin/pip
user@ubuntu16:~$ pip install -U ggplot
Collecting ggplot
[...]

PR #186, opened 2017-04-28
Installation with ggplot

tmorrell:
Currently ggplot is not compatible with python3.6 that is installed with new versions of anaconda.  While there is a solution: https://github.com/datacarpentry/python-ecology-lesson/issues/164, it involves downloading a ton of additional packages (which is difficult to do during a workshop with poor wifi).  I have confirmed that a standard install of Anaconda3-4.2 is compatible with the conda-forge version of ggplot, and think this is the easiest installation solution for now.  I've created a separate issue to discuss improvements to the install process.

I have not been able to find a comparable solution for miniconda, so I included the environment instructions.  This could use more testing.

I also switch contexts from bokeh to conda-forge.  The bokeh version of ggplot is older and does not support the line plot example.

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -57,8 +57,8 @@ conda install -c conda-forge ggplot
 
 ### Miniconda installation
 
-Miniconda is a "light" version of Anaconda that you can install instead. 
-If you install and use Miniconda you will also need to install the workshop packages.
+Miniconda is an alternative "light" version of Anaconda. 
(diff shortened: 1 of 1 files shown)


wrightaprilm:
These instructions work for me. I'm always a little hesitant to merge install instruction changes without anyone else trying them. @qjcg, would you mind double-checking them?

PR #189, opened 2017-05-01
simplify commands, clarify slicing, remove index

tmorrell:
Add small clarification for negative index slicing.  Simplify data subset by using 'tail' instead of the confusing all (except last element) slice.  Makes the re-import correct by removing the extraneous index column.


qjcg:
Looks good, thanks for this @tmorrell !

PR #192, opened 2017-05-31
Add section on creating tables to SQLite episode.

godfoder:
It seemed to me like the SQLite section was missing a key interaction with the database. It had no information about modifying or creating tables in SQLite. Hopefully this short example should give attendees enough of guide to show them its possible and an example of one way to do it.


wrightaprilm:
This looks good to me. Is this for instructor training? If so, I'll tag it before merging and closing.

godfoder:
This is for checkout.

wrightaprilm:
OK, thanks for contributing. I've tagged it as instructor-training and I'm merging it now.

Issue #190, opened 2017-05-05
Challenge Statistics nunique() error


nerdcommander:
in the "Challenge - Statistics" section, item #2 asks "What is the difference between len(plot_names) and plot_names.nunique()?"  

I'm assuming that the point is to demonstrate that there is often more than one way to accomplish this (and most tasks) in Python and programming.

the potential problem here is that  the variable `plot_names` contains an array, and the `nunique()` method does not work on an array, only on a pandas series, so you don't get the number of unique items, you get an error.

`plot_names.nunique()` throws an error - `AttributeError: 'numpy.ndarray' object has no attribute 'nunique'`
`surveys_df.plot_id.nunique()` works since the first part produces a series which then works for `nunique()`.

unless the point is to get this error, then my apologies.

---

Thanks for contributing! If this contribution is for instructor training, please send an email to checkout@carpentries.org with a link to this contribution so we can record your progress. You’ve completed your contribution step for instructor checkout just by submitting this contribution.  

Please keep in mind that lesson maintainers are volunteers and it may be some time before they can respond to your contribution. Although not all contributions can be incorporated into the lesson materials, we appreciate your time and effort to improve the curriculum.  If you have any questions about the lesson maintenance process or would like to volunteer your time as a contribution reviewer, please contact Kate Hertweck (k8hertweck@gmail.com).  

---

andreas-h:
can this be closed? seems to be fixed by #198

wrightaprilm:
You are correct, @andreas-h. I entirely forgot this was still open. Thanks all!

PR #198, opened 2017-07-11
Adapt instructor notes and 01 episode

stijnvanhoey:
For the 01-starting-with-data.md episodes, following adaptations are suggested:
- renamed the variable name used to define the dataframe.groupby object from `sorted_data` to `grouped_data`
- As defined in #190, the current solution in the instructor nodes are confusing. Actually, in the challenge description the usage of `nunique` on the variable `plot_names` does not make much sense. So instead, this is applied to the data frame column instead of the `plot_names` variable
- The syntax of the `total_count` is rather confusing, as the `surveys_df['plot_id']` is used inside the `groupby` operation: `total_count = surveys_df['record_id'].groupby(surveys_df['plot_id']).nunique()`. This is adapted to `total_count = surveys_df.groupby('plot_id')['record_id'].nunique()` which is in line with the earlier code.
- the *make a list* challenge has been moved out the basic math functions section, as this has nothing to do with a basic math operation, but is in line with the previous section

### DIFF
--- a/_episodes/01-starting-with-data.md
+++ b/_episodes/01-starting-with-data.md
@@ -413,6 +413,13 @@ Or, we can also count just the rows that have the species "DO":
 surveys_df.groupby('species_id')['record_id'].count()['DO']
 ```
 
+> ## Challenge - Make a list
+>
+>  What's another way to create a list of species and associated `count` of the
+>  records in the data? Hint: you can perform `count`, `min`, etc functions on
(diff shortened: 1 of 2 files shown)


wrightaprilm:
This looks great, thank you! Is this for instructor training check out? If so, I'll tag it as such.

stijnvanhoey:
Thanks! Is indeed for the instructor traingin checkout.

Issue #200, opened 2017-07-17
Character data type in Episode 03


yeemey:
In the episode on data types and format, the data type for strings is referred to as "character." It would be more consistent with the [Python docs](https://docs.python.org/3/library/stdtypes.html#text-sequence-type-str) to rename this as text or text sequence data type. It will also be less confusing, given the introductory sentence of the "Character data type" section: 

> Strings, known as Objects in Pandas, are values that contain numbers and / or characters.

wrightaprilm:
Sure thing. Is this for instructor training checkout (i.e., are you going to submit a pull request making this change, or would you like me to handle it)?

yeemey:
I will make this change, and submit a pull request, thanks!

yeemey:
Yes, this is for instructor training checkout.

wrightaprilm:
Addressed by #201. Thank you!

PR #201, opened 2017-07-18
Renamed character dtype to text dtype.

yeemey:
Resolve issue #200


wrightaprilm:
I'm happy with this. I'll merge and tag as instructor checkout. Thank you!

PR #202, opened 2017-07-25
Mentioned the ability to use LaTeX notation in legends and axis labels

bgbg:
Scientists often operate on mathematical equations. Being able to use them in their graphics has a lot of added value. The provided example is an only slight modification to the original one. Nevertheless, it produces a significantly better result.

### DIFF
--- a/_episodes/07-putting-it-all-together.md
+++ b/_episodes/07-putting-it-all-together.md
@@ -214,25 +214,8 @@ several times:
 The function `legend()` adds a legend to the figure, and the optional keyword
 arguments change its style. By default [typing just `plt.legend()`], the legend
 is on the upper right corner and has no shadow.
-
-Scientists often operate on mathematical equations. Being able to use them in their graphics has a lot of added value. Luckily, Matplotlib provides powerful tools for text control. One of them is the ability to use LaTeX mathematical notation, whenever text is used (you can learn more about LaTeX math notation here: https://en.wikibooks.org/wiki/LaTeX/Mathematics). To use mathematical notation, surround your text using the dollar sign ("$").  LaTeX uses the backslash character ("\") a lot. Since backslash has a special meaning in the python strings, you should replace all the LaTeX-related backslashes with two backslashes. 
  
- ```python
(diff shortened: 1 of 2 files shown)


wrightaprilm:
I'm of slightly two minds about this: it's very late in the lesson to introduce a new concept. But I do think that being able to put real, complex mathematical descriptions of what you're plotting is important. This lesson also generally needs a rewrite in light of how we've restructured the plotting. 

I think I'd still like to merge this for now. Can I get you to rewrite the paragraph you've submitted to emphasize the utility of this? The majority of Data Carpentry learners will not be folks who have Latex experience. 

Thanks for submitting this - is it for instructor training?

bgbg:
Yes, this is for instructor training.

Also, I spelled the motivation out, before showing the new example.

wrightaprilm:
Sorry for being so late on this; moving. 

I thought about this a little bit more. The paragraph looks good, but it's quite late in the lesson to introduce a new functionality, particularly something as complex as latex. So I think the way to approach this would be to create a link in the lesson noting that more complex mathematical notation is possible, but including the actual text of the pull request and the example in the instructor's guide. That way, instructors working with more advanced classes can pull the material in, but those working with novices can note the functionality and not go into it.

How does that sound?

bgbg:
This makes sense. Can you point me to the right place to which this should
be moved? Also, since we're moving it to an "advanced" section, is it a
good idea to add an even more complicated example?

wrightaprilm:
So sorry to be late getting back on this - started a new faculty job 2 weeks ago. The instructor guide is here: https://github.com/datacarpentry/python-ecology-lesson/blob/gh-pages/_extras/guide.md

There's not yet a section for lesson 07, so go ahead and add one. If you want to just include the simple example, that's fine, but if you have another example in mind, you can include that, too. That way the instructors have a little leeway on what they want to incorporate and what works for their audience.

Thanks!

bgbg:
It's my turn to apologize. I will probably get to this task in a week or two

wrightaprilm:
This looks great, I'll go ahead and merge this. Thanks for all your work on this one!

PR #204, opened 2017-08-15
Update 05-loops-and-functions.md

LikeFokkens:
Removed 2. from  Challenge - Modifying loops (and renumbered subsequent challenges). The reason to remove it is that it refers to 'range' while the range function has not been explained in any of the previous episodes.


wrightaprilm:
This looks good, I'll merge it now. Thanks! Is this for instructor training? If so, I'll tag it as such.

LikeFokkens:
Hi April,

Hi April

Yes it is for instructor training, thank you.

cheers
Like

April Wright schreef op 28-08-2017 21:26:

PR #209, opened 2017-09-05
Instructor checkout - added keypoints to episodes 00-08, added questions to episode 08

LucaDiStasio:
Submission for the Data Carpentry Instructor Checkout.

The following changes have been made:
- key-points have been added to episodes from 00 to 08
- questions have been added to episode 08


wrightaprilm:
It looks like everything here is covered in the other two pull requests, so I'm going to close this one. If that's not accurate, let me know and I'll reopen.

LucaDiStasio:
No, that's right. I messed up somewhere when pushing to branches, I guess.
I was trying to separate the issues in different branches, but two ended up
identical. I pushed it because I wanted a double check from another person.

So, in summary:
Yes, it's should be closed
And thank you for checking it out!

Best,
Luca

PR #215, opened 2017-10-04
Add variable assignment to example.py

GeneMachines:
Please delete the text below before submitting your contribution. 

The example.py script in the first lesson, 00-short-introduction-to-Python.md, which demonstrates the print function might be confusing as it would not work if run as is. The script was altered so that assignment of the string "Data Carpentry" to the variable "text" was made in the script.

### DIFF
--- a/_episodes/00-short-introduction-to-Python.md
+++ b/_episodes/00-short-introduction-to-Python.md
@@ -83,7 +83,7 @@ However, in a script, a `print` function is needed to output the `text`:
 ```python
 # A Python script file
 # Comments in Python start with #
-# The next line assigns the string "Data Carpentry" to the variable `text`.
+# The next line assigns the string "Data Carpentry" to the variable "text".
 text = "Data Carpentry"
 # The next line uses the print function to print out the text string
 print(text)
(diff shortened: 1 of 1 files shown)


wrightaprilm:
This looks good to me. Look alright to you, @pdoehle?

pdoehle:
Hi @wrightaprilm,

Sorry for the delay in my response. `text = "Data Carpentry"` was defined two code snippets earlier. I felt that the example made sense working through the lesson. Do you think redefining the variable again would get too cluttered, or brings clarity to the example?

GeneMachines:
Hi @pdoehle, I'll clarify my thinking a bit : ) 

I follow your logic, and I think anyone with experience scripting will understand the example as written, but I think a total beginner might get confused by the structure. For example, the student might imagine that by opening an interpreter and setting a variable that the variable persists after the interpreter is closed. This might be seen as similar to when you set a variable in the shell and can later reference it in a script from the same console. While more verbose, I felt this change clarifies the example for people with a very limited understanding of programming.

pdoehle:
Hi @GeneMachines,

Ah, ok that makes sense to me, thanks for the clarification. Sounds good! @wrightaprilm, what is our procedure for merging?

wrightaprilm:
You go ahead and merge this one. Just click the merge button. Normally, I'd close the discussion at the same time, but I have one last question that needs answering - is this for instructor training, @GeneMachines? If we get an affirmative, @pdoehle, on the right hand side under "Labels", click and type in "Instructor Training" to tag it as such.

pdoehle:
Thanks @wrightaprilm, I'll keep an eye on it and close everything out when we hear back.

PR #216, opened 2017-10-25
tip & note added to 02-index-slice-subset

brownsarahm:
tip about using .head() to control output throughout 02 as comment tip in lesson and as a more extended note in the instructor guide

Also in 02 a note on what in the error traceback indicates that the error is


wrightaprilm:
Looks great, thank you for contributing! If this looks good to you, @pdoehle, I'll go ahead and merge.

PR #222, opened 2017-11-06
Introduce type function earlier + some wording changes

colinmorris:
The main goal in this P.R. was to introduce the `type()` function earlier (in the 1st lesson rather than the 2nd). It's a pretty useful tool for Python learners, and the idea of types is already being discussed in the 1st lesson.

I also wanted to remove the references to `__class__` from the second lesson. I think it's probably too advanced and doesn't add anything beyond `type`.

The rest is mostly incidental wording changes I made while I was in the neighborhood, which I hope are improvements.

### DIFF
--- a/_episodes/00-short-introduction-to-Python.md
+++ b/_episodes/00-short-introduction-to-Python.md
@@ -56,8 +56,7 @@ Hello World
 
 ### Strings, integers and floats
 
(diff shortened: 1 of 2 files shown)


wrightaprilm:
This looks good. I resolved a conflict in here, so if you'd like to have a last look over everything before I close the PR, have at it. 

Thanks!

Issue #225, opened 2017-11-07
'If loop' is potentially confusing terminology


colinmorris:
It's used throughout [lesson 05](http://www.datacarpentry.org/python-ecology-lesson/05-loops-and-functions/). 

I think the word "loop" might give the impression that these things go around and round like for loops, executing more than once.

I'd suggest renaming to "if statement" (or "conditional", or "if/else").

tdpearson:
I agree and have created the above PR.

wrightaprilm:
Solved by #227.

PR #227, opened 2017-11-08
replace 'if loops' statements

tdpearson:
This is in regards to issue #225.

### DIFF
--- a/_episodes/05-loops-and-functions.md
+++ b/_episodes/05-loops-and-functions.md
@@ -504,11 +504,11 @@ doesn't correspond to any existing object.
 > those with default values in the function definition!)
 {: .challenge}
 
-## If Loops
+## If Statements
 
-The body of the test function now has two conditional loops (if loops) that
-check the values of `start_year` and `end_year`. If loops execute the body of
-the loop when some condition is met. They commonly look something like this:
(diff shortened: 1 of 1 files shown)


_episodes/05-loops-and-functions.md
```diff
@@ -538,11 +538,11 @@ Which would return:
 Change the value of `a` to see how this function works. The statement `elif`
 means "else if", and all of the conditional statements must end in a colon.
 
-The if loops in the function `yearly_data_arg_test` check whether there is an
+The if statements in the function `yearly_data_arg_test` check whether there is an
 object associated with the variable names `start_year` and `end_year`. If those
-variables are `None`, the if loops return the boolean `True` and execute whaever
+variables are `None`, the if statements return the boolean `True` and execute whatever
```
colinmorris: Not a problem introduced in this P.R., but looking at this I'm noticing some problems with the wording. I think what is meant here is that if the variables are `None`, the if *condition* (i.e. the thing between `if` and the colon) *evaluates to* the boolean `True`.

It's a small difference, but I think it's important to reserve the verb "return" for talking about the results of functions/methods.

wrightaprilm:
Looks good to me, thanks!

Issue #221, opened 2017-11-05
Bug BBQ: Data setup confusion


colinmorris:
The "Data" section of the [setup document](http://www.datacarpentry.org/python-ecology-lesson/setup/) says:

> For this lesson, we will use two files for the data. Download these files to your computer. Once you click on them they will be automatically downloaded to your default download directory.

I see a few problems here.

1. There are actually 7 files listed after this, not 2.
2. It is not true that clicking each link will initiate a download. This happens with the figshare links, but the links to the github-hosted files bring up a web preview. (To download the file, you need to click the "raw" button, then "save page as").
3. There doesn't seem to be any guidance about about making these files findable in the Python environment that the students set up.

1 and 2 are pretty minor problems, but I think 3 is something that students could easily get stuck on.

If I just follow the setup instructions, the first time I call `pd.read_csv("surveys.csv")` in the [Starting With Data](http://www.datacarpentry.org/python-ecology-lesson/01-starting-with-data/) lesson, I'm going to get an error, because Python is running from my home directory (or whatever the pwd happened to be when I ran `jupyter notebook`) and the csv files are in my Downloads directory.

Also, it's worth pointing out that some lessons even make different assumptions about the relative path to the data. e.g. the [Data Types and Formats](http://www.datacarpentry.org/python-ecology-lesson/03-data-types-and-format/) lesson has the line `surveys_df = pd.read_csv("data/surveys.csv")`.

## Suggested solution

I'd suggest adding a section to the setup document (before the "Data" section) about setting up a working directory. There's a good example of this in the [R ecology lesson](http://www.datacarpentry.org/R-ecology-lesson/00-before-we-start.html#organizing_your_working_directory).

Lessons/setup should be amended to clarify that...
- when starting the interpreter or the notebook server, students should be in the working directory they created during setup.
- data files should be saved to the `data/` subdirectory of their working directory

And paths to data files in the code examples should be made to consistently use the `data/` prefix.

Another minor suggestion: It would be a nice little time-saving measure to host a zip file containing all 7 data files, just to save some clicking.

cmaimone:
One way to provide the ability to download the files as single zip file, while not creating issues with keeping the zip file updated as the data files change, is to use [DownGit](https://minhaskamal.github.io/DownGit/#/home).  

[...]

PR #231, opened 2017-12-06
Updated data file download links in setup

cmaimone:
Related to issue #221 but doesn't resolve it.  

Added option to download add data files from a single zip file.  Also updated all data file urls to point to the data directory for the repository, to 1) be consistent (for any future updates), and 2) make downloads automatic (previous links were to github pages, not raw files).


wrightaprilm:
Looks great, merging it now. Thanks!

PR #233, opened 2017-12-13
I searched for and corrected paths to data files in lessons. 

joshsteele:
This pull request references part of issue #221. I have gone through and corrected and checked the paths to the data files in the lessons. I did not check to make sure that the downloaded data files were placed into a folder named "data". I think this is the case, however.  I also cleaned up a formatting issue where a comment in a python code section was reading as a heading. There were a few weird formatting things I tried to clean up in along the way. Feel free to keep them if they're helpful or leave them if they're not.

### DIFF
--- a/_episodes/01-starting-with-data.md
+++ b/_episodes/01-starting-with-data.md
@@ -129,7 +129,7 @@ an element in the data structure.
 
 ```python
 # note that pd.read_csv is used because we imported pandas as pd
-pd.read_csv("surveys.csv")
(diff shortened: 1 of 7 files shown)


wrightaprilm:
Really nice work. I'll got ahead and merge this. You are correct that we don't specify where to put the data, so I'm going to leave issue #221 open for now, since there are more instructor training folks checking out soon and that's a nice checkout project.

Thank you!

PR #234, opened 2017-12-28
Setup alternative for mac os

guoxiaohu:
For mac os,  I think using homebrew is much more convenient than using anaconda to install workshop dependencies. I haven only tried on the high Sierra though

### DIFF
--- a/_extras/guide.md
+++ b/_extras/guide.md
@@ -6,10 +6,10 @@ permalink: /guide/
 
 # Challenge solutions
 
-## Install the required workshop packages on Mac OS 
+## Install the required workshop packages on Mac OS with Homebrew
 
 
-Here we are using one of the popular package manager called Homebrew.
+Homebrew is one of the popular package managers used by many Mac users.
 
 To install Homebrew, you need have xcode command line tools installed, 
 
--- a/_extras/guide.md
+++ b/_extras/guide.md
@@ -6,6 +6,49 @@ permalink: /guide/
 
 # Challenge solutions
 
+## Install the required workshop packages on Mac OS 
+
+
+Here we are using one of the popular package manager called Homebrew.
+
+To install Homebrew, you need have xcode command line tools installed, 
+
+from the terminal, type:
+```
+xcode-select --install
+```
+
+then
+```
+/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+```
+
+Run the following command to ensure Homebrew is installed properly:
+
+```
+brew doctor
+```
+
+install python 3:
+
+```
+brew install python3
+```
+
+All the rest of workshop packages can be installed by pip3 which comes together with python 3:
+
+```
+pip3 install numpy pandas matplotlib jupyter
+pip3 install bokeh ggplot
+```
+
+pip3 can automatically figure out all the dependencies that required by the above packages.
+
+Launch a python command prompt, type:
+```
+python3
+```
+
 ## Checking installations. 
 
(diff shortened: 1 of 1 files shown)


wrightaprilm:
Hello-

I'm hesitant to merge this for a few reasons.

1. It's something additional for me to be checking (in case an update breaks the installs) when I'm maintaining alone.
2. It brings us out of alignment with the SWC installs. So if an instructor is using lessons from both, we don't know that they work.
3. It's something different for our new instructors to learn to be managing  when they're already busy.
4. The XCode install is much larger (4 times larger than Anaconda, 6 larger than Miniconda?), which complicates troubleshooting if someone's preworkshop installation did not go as planned.

So I'm not opposed to this, but I do have some concerns that need some discussion. (cc @ErinBecker - is there anyone you'd recommend looping in who thinks a lot about installation issues?)

guoxiaohu:
Hello,
   Thanks for pointing this out, as this is just alternatives for mac users, so my intention is for those who has homebrew installed and don't have to download big file again as you have already pointed out.

  Fair point for comment 1 !
  I can't comment on the comment 2 as I don't know much about other SWC installs.

   for the comment 3, I don't think it would make that much effort to use pip3 manage python distribution instead of conda.

   for the comment 4, we only need "xcode command line tools", it's part of xcode but much smaller than xcode itself, it take only few minutes to download and install.

Thanks !

naupaka:
Some thoughts:

*I should note at the beginning that I'm from the R side of the fence, and am not a python expert by any stretch, but I have been through many learner laptop workshop setup struggles.*

I'm -1 on merging this. I use homebrew all the time and agree that it's lovely (I even donate to homebrew on Patreon because it really is a huge part of what makes open source software on macOS work so well), but I think we don't want to require another installer (and a platform-specific one at that) for Carpentry workshops. 

As much as possible, we want everything to be the same across all platforms Windows/macOS/Linux. Anaconda works on all platforms. Additionally, anaconda offers more than just package installation -- conda environments are a powerful/useful thing to have available, especially down the road as people's skills mature, and they are not provided in any form by homebrew itself.

Additionally, I have had my share of struggles with package installation problems caused by the homebrew python and packages conflicting with pre-installed macOS versions. Isolated conda environments are just much easier to turn on and off as needed and are more flexible if you need different versions of python or of particular packages. 

If anything, I might stick something like this on the instructor notes page, for times when anaconda installation fails for some reason. Generally speaking, though, I think there are a lot of advantages to keeping things as standardized as possible for workshops in particular (e.g. why we ask everyone, including instructors, to use `nano` instead of letting people use a variety of other editors). It makes it easier for novices, for helpers, and for instructors if they can make the general assumption that everyone is on the same page...

I do, however, think that it is definitely worth pointing out (if it isn't mentioned elsewhere already) that mac users with certain OS versions may need to run `xcode-select --install` even apart from any homebrewing, in order to have access to `git`.

wrightaprilm:
I think putting it in the Instructor Guide is a good compromise. That way, it's accessible for self-study students, or instructors embedding the material in larger classes, which I think using HomeBrew would be really great for!  Sound good?

guoxiaohu:
I think I totally agree @naupaka to keep single virtual env across all OS. Put these into Instructor Guide make sense to me. However, where is the instructor guide repo ? I can't find the link ?

wrightaprilm:
We keep it here: https://github.com/datacarpentry/python-ecology-lesson/blob/gh-pages/_extras/guide.md

There's a section on installs, where a discussion of alternative package managers could fit cleanly.

wrightaprilm:
Looks great, merging now! Thank you!

PR #235, opened 2018-01-18
Adding an explanation of CSV

maxvillev:
It would be good to help newbies get past the acronym, and to know the benefits and variations of delimited text files. I've added four more sentences to explain. It's not at the first reference to CSV, but to do it earlier might have been distracting.

### DIFF
--- a/_episodes/01-starting-with-data.md
+++ b/_episodes/01-starting-with-data.md
@@ -115,7 +115,7 @@ time we call a Pandas function.
 
 # Reading CSV Data Using Pandas
 
-We will begin by locating and reading our survey data which are in CSV format.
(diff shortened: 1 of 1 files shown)


wrightaprilm:
Thanks for this. The materials assume that people have come to the lesson as part of a workshop, which is not always the case. Would you mind adding in [a link ](http://www.datacarpentry.org/spreadsheet-ecology-lesson/05-exporting-data/)to the DC-spreadsheet lesson on CSV files? That way, if people need more info, or to find out how to even get their data into CSV, they have the resources to do that?

wrightaprilm:
Looks great, merging now. Thank you @maxvillev!

PR #237, opened 2018-01-24
Calculating more meaningful statistics by grouping

daviddesancho:
Averages on weights across species is not very informative, while
averages across genres for each of the different species may be more
sensible. Hoping this will help users see the profit you get from
using Pandas.

### DIFF
--- a/_episodes/01-starting-with-data.md
+++ b/_episodes/01-starting-with-data.md
@@ -335,7 +335,7 @@ can quickly calculate summary statistics by a group of our choice.
 
 ```python
 # Group data by sex
-grouped_data = surveys_df.groupby('sex')
+grouped_data = surveys_df.groupby('species_id')
 ```
 
 The **pandas function `describe`** will return descriptive stats including: mean,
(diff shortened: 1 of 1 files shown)


wrightaprilm:
I actually think this makes the output look a little more messy. I had a look at what R-ecology is doing recently, and it looks like they are grouping on both sex and species-id (i.e., in pandas: grouped_data = surveys_df.groupby(['species_id', 'sex'])). What would you think about moving to an example where we group on both?

daviddesancho:
I agree that the output is more messy (or, well, longer), but it however makes more statistical sense. Remember you are calculating a mean value, which is a sensible statistical descriptor for normally distributed data. That's probably not the case for my proposal, i.e. for datasets of species, as they would include both male and female. However, the assumption of the normal distribution is more evidently wrong when you are averaging over gender (i.e. including males of different species and then naively calculate a mean value of, for example, their weight).

What you suggest would in fact that would be a sensible replacement to point 2 in the "Challenge" that comes right after my suggested change. There you would do the `surveys_df.groupby(['species_id', 'sex']))` as opposed to the current `surveys_df.groupby(['plot_id','sex'])` grouping.

trallard:
I'd be tempted to follow @wrightaprilm 's suggestion. Leave the current grouping strategy in the main lesson content to keep the output concise and clear and then follow with `surveys_df.groupby(['species_id', 'sex']))` in the challenge afterwards. 

I know other approaches would make more sense from a statistical point of view but this solution keeps the lesson neat and then introduce the more sensible `surveys_df.groupby(['species_id', 'sex']))`

maxim-belkin:
Closing this for now. I agree that meaningful examples would be better though. So, perhaps, there should be a sentence that would say that this is an example of not-very-useful statistics?

Thanks @daviddesancho anyway!

PR #240, opened 2018-01-30
Formatting for the columns of data

hartmanbaker:
Including information about formatting at the beginning of exercise 1 in case other users don't see all 9 columns of data. The first time I tried this exercise, that threw me off for a minute.

### DIFF
--- a/_episodes/01-starting-with-data.md
+++ b/_episodes/01-starting-with-data.md
@@ -230,8 +230,9 @@ Never fear, all the data is there, if you scroll up. Selecting just a few rows, 
 easier to fit on one window, you can see that pandas has neatly formatted the data to fit
 our screen:
 ```python
-# the [x:y] appended to the variable name selects a slice of the data, columns x through y-1
->>> surveys_df[5:10]
+
+>>> surveys_df.head() # The head() function displays the first several lines of a file. It
+		      # is discussed below.
    record_id  month  day  year  plot_id species_id sex  hindfoot_length  \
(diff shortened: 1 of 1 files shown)


wrightaprilm:
I conceptually like this a lot, particularly since these materials get used a lot for self-study, so an individual learner might not know what's up. I made one note in review, to swap out the indexing for the .head() function. We cover that function almost immediately after in the "Useful Ways of Viewing Data" section, so I think that keeps the whole discussion a little tighter.

_episodes/01-starting-with-data.md
```diff
@@ -173,7 +173,79 @@ object by typing its name into the Python command prompt.
...
+35548     NaN  
+
+[35549 rows x 9 columns]
+```
+Never fear, all the data is there, if you scroll up. Selecting just a few rows, so it is
+easier to fit on one window, you can see that pandas has neatly formatted the data to fit
+our screen:
+```python
+# the [x:y] appended to the variable name selects a slice of the data, columns x through y-1
+>>> surveys_df[5:10]
```
wrightaprilm: Could we swap out the index for surveys_df.head()? This command is discussed in detail in the next section, so we can keep all of this fairly compact.

hartmanbaker: That sounds great! Just something to demonstrate that this view is expected behavior and not a mistake.

wrightaprilm: OK, I'll make the edit and merge.

PR #247, opened 2018-03-05
Changing the mean weight to the sum of all weights

mboisson:
The mean weight across various species is rather meaningless. By changing it to the sum of the weights, we can relate this to the biomass of the plot 
https://en.wikipedia.org/wiki/Biomass_(ecology)

This makes the exercise more meaningful.

### DIFF
--- a/_episodes/01-starting-with-data.md
+++ b/_episodes/01-starting-with-data.md
@@ -528,7 +528,7 @@ total_count.plot(kind='bar');
 
 > ## Challenge - Plots
 >
-> 1. Create a plot of average weight across all species per plot.
+> 1. Create a plot of sum of the weights across all species per plot. This is called the biomass for a given plot.
 > 2. Create a plot of total males versus total females for the entire dataset.
 {: .challenge}
 


wrightaprilm:
Sorry for being so late getting back to this. I'm a little hesitant to make changes that diverge the R and the python lessons unnecessarily. How would team R (@fmichonneau @aurielfournier ) feel about this change?

fmichonneau:
I think the only time in the lesson we do something like this is in questions 4 of this challenge: http://www.datacarpentry.org/R-ecology-lesson/03-dplyr.html#challenge19

I'm open to making it more biologically relevant, but it would require refactoring the entire challenge...

aurielfournier:
I agree with @fmichonneau

wrightaprilm:
I could have sworn the R lesson did averages across species across plots in the Starting With Data lesson, but it looks like you aggregate the number of males and females caught across plots. Is that right (I know our templates are organized a little differently)? If so, I might create an issue for us to move to that example so the tasks between the two sets of lessons are broadly similar.

Issue #246, opened 2018-03-03
Bug BBQ: ggpy or plotnine when introducing grammar of graphics - Discussion


stijnvanhoey:
The current lesson [Making Plots With ggplot](http://www.datacarpentry.org/python-ecology-lesson/06-visualization-ggplot-python/) is using the package [ggplot/ggpy](http://ggplot.yhathq.com/). However, when checking the repository for recent developments, it says `Latest commit b6d23c2  on Nov 20, 2016`. At the same time, the [plotnine](https://github.com/has2k1/plotnine) package is still under active development and provides the same port of ggplot (grammar of graphics) for Python users. I've had good experience with plotnine. 

Would it be a good idea to switch and change the content towards the implementation of the `plotnine` package? I'm willing to support the conversion (cfr. we already did [something similar](https://github.com/jorisvandenbossche/DS-python-data-analysis/blob/master/_solved/visualization_02_plotnine.ipynb) during a recent course), but want to check if this is worthwhile?

stijnvanhoey:
Just a a test, I adapted the first section [plotting-with-plotnine](https://github.com/stijnvanhoey/python-ecology/blob/plotnine-conversion/_episodes/06-visualization-ggplot-python.md#plotting-with-plotnine) and the syntax does not seem to be very different. Although, as mentioned in #226 , the next sections maybe require some adapations anyway?

wrightaprilm:
This is a really good question that I'd like some broader opinions on. We have the bug bonanza coming up, so now is a good time to have it. For context: the plotting lesson has always been a bit of a beast, and a lot of that is due to the fact that there's more diversity in what people do for plotting in python, compared to R. There's a fork of this repo that gets used in a semester-long course, doing plotting in Seaborn, for instance. 

We switched over to ggplot a bit ago from MatPlotLib, for greater continuity with the R lesson. And that's something I had mixed feelings on at the time. Having the continuity is nice, since learners might use multiple lesson sources. But MatPlot is part of the NumFocus ecosystem, and has a sustainability plan, etc. Basically, heading off the issue we see now with ggplot, where there is not active development happening, and a change to Pandas or python that breaks the interface might not be handled in a timely manner. 

There's pros and cons here. I prefer the idea of active maintenance, for responsiveness to change. But I'm also hesitant to have the community sink the time into switching if we end up in the same place - all set up with a new package that drops out of development.

I'd like to throw out a few questions for discussion as we prep for new lesson releases:

[...]

Issue #226, opened 2017-11-07
Pandas code in episode 06 (visualization) seems very advanced


colinmorris:
(Sorry for the flurry of issues/PRs. I'm preparing to do a demo for this lesson, so I've been poring over it for the last day or two.)

I use Pandas pretty regularly, but I was a little thrown by some of the code in this episode until I stepped through it in the interpreter. Particularly:

```python
yearly_counts = surveys_complete[['year','species_id']].groupby(['year', 'species_id']).size().reset_index()
yearly_counts.columns = ['year','species_id', 'n']
yearly_counts
```

```python
yearly_sex_counts = surveys_complete.groupby( ['year','species_id', 'sex']).count()
yearly_sex_counts['n']  = yearly_sex_counts['record_id']
yearly_sex_counts = yearly_sex_counts['n'].reset_index()
yearly_sex_counts
```

As far as I can tell, the following are all new at this point in the lesson:
- `reset_index()` (it's briefly glossed over in the Merging lesson, though in that case it's used to a very different effect)
- renaming the columns of a DataFrame
- adding a new column to a DataFrame

Are students expected to understand this code? Or is the idea to focus on teaching the visualization stuff, and tell students they don't need to worry about the code that generates the data to be plotted?

There's a lot to unpack to really explain what the `reset_index` does and why it's necessary - you need to know about Series and multi-level indexes.

Also, minor issue: the first example uses `.size()` and the second uses `.count()`. I'd suggest sticking to one of them (the groupby material in episode 01 introduces `count()`, so I'd go with that one). Or if we're deliberating taking advantage of their [differences](https://stackoverflow.com/questions/33346591/what-is-the-difference-between-size-and-count-in-pandas), that's worth calling out.

PR #248, opened 2018-03-07
Plotnine conversion of the plotting episode

stijnvanhoey:
As currently discussed in #246, the `ggpy` package currently used in the plotting episode provides a Grammar of Graphics implementation, but seems still unstable and no longer actively maintained. As the syntax of `plotnine` is very similar and the package actively maintained, the conversion seems valid.

Even though I do understand the conversion is maybe a decision that need to be made broader, the main reason I'm putting this PR already out is the structure and rendering of the current episode. A challenge is currently a title (actually twice), a solution spread over 5 code blocks without any textual clarification,...

Furthermore, I did my best to adapt the flow of the episode to counteract the current 'advanced' feeling as reported by #226 @colinmorris maybe you could review this with respect to your issue?


has2k1:
For an alias of `plotnine`, you should consider `p9` instead of `pn`. I have seen `plotnine` abbreviated that way in a couple of places.

stijnvanhoey:
Thanks for the suggestion, I'll convert the aliases!

wrightaprilm:
I'm not going to get a chance to look in any depth at this  pull request until next week. I'm going to see if I can solicit some extra hands-on to take a peek both at this and the wider discussion surrounding plotting, as well. One feasible thing might be to put the lesson using p9 on a separate branch that instructors could check out from.

trallard:
@wrightaprilm do you need help with this? I can have a look at this PR over the BBQ bonanza

wrightaprilm:
@trallard Absolutely, more eyes on this, the better!

wrightaprilm:
All right, this passes checks on my Mac and PC. I'm going to merge, then file a new issue asking for folks to check it. Thanks so much for your work on this lesson!

colinmorris:
Sorry, I'm late to the party on this PR, but I just wanted to say that, in terms of the pandas complexity I brought up in #226, I think this is a huge improvement!

Just to demystify it even further, my one suggestion would be to show excerpts of the dataframes/series you're getting back when you do the `yearly_counts` stuff. I filed an overall issue about showing outputs here: #268.

PR #250, opened 2018-03-13
corrected dataframe output descriptions

oxpeter:
the surveys_df data have 35,549 rows, not 33,549 as written.

Also, the graph shown from the png file is the number of counts for each
species, not the weight of each species, as the description and
filename would indicate. png file has been duplicated and renamed
to avoid confusion. Original 'weightBySpecies' has not been
removed until it can be determined it is not referenced by another
link somewhere.

link in lesson 01 updated to point to new image.


wrightaprilm:
Great, thank you!

PR #256, opened 2018-04-07
remove redundant code segment and white spaces

eotp:
There is no need to show the function `yearly_data_arg_test` once again; there are basically no changes compared to the function defined above. Moreover, in the body the condition is checked via `if not`, and not with `if is None` as above. Though both functions yield the same result in the present use case, it is probably more confusing than helpful to provide two different ways for checking the condition.


wrightaprilm:
Right on, thank you very much!

PR #257, opened 2018-04-08
new introduction

katrintirok:
Python Ecology lesson

I am teaching both R and Python for Ecology and I found the R lesson flows a bit better than the python lesson and I think both lessons could be synchronised better to ensure, in terms of datacarpentry, they teach the same content and also to make it easier for instructors to switch between teaching R and Python. 

This mainly touches on the structure of the lessons.
I found that the python lesson missed a solid introduction for novices to programming (and python), so I started drafting a ‘Before we start episode’ which pretty much resembles the one in the R lesson. I wrote it for Spyder but it should be fairly easy to add a similar section for notebooks so instructors can choose which one to teach (this makes me think that a notebook section could also be added to the R lesson in return). 

Then I also rewrote the ‘Intro to python’ episode, also to make it closer to the R lesson and to introduce some really basic concepts (like libraries, data types) in a more structured way. With that, content from existing episode ‘data-types-and-formats’ is merged into ‘intro’ episode.

I am also busy reviewing the ‘starting-with-data’ episode accordingly … (not in this pull request).

In this pull-request the branch new-intro includes new episode ’00-Before-we-start’ and ’01-Intro-to-python’.

These could use some proof-reading. I am happy to do more changes to the two episodes if necessary.

Also, it may solve some of the open issues, I added comments on some open issues accordingly.


wrightaprilm:
Hi Katrin-

Thanks for your contribution! We're having a Bug BBQ at the end of this week, focusing mostly on shoring up existing materials, but that also makes it a nice time to evaluate new contributions, since we get more eyes on the repository than average. Would you mind splitting this PR into two, one for the pre-lesson and one for the rewritten first lesson?

katrintirok:
thanks, I will split it up. Yes, I wanted to get it out before the Bug BBQ to get some more feedback.

katrintirok:
I think this pull-request can be closed, one part was merged and the other part is in #260

PR #258, opened 2018-04-08
Update starting with data

katrintirok:
Update starting with data:
- updated link to surveys.csv to get new release of data on figshare
- removed redundant information regarding libraries etc. handled in rewritten intro-to-python episode (#257 )
- added bits of info regarding Spyder usage vs. Notebook usage
- changed first challenge
- changed 'Basic Math Functions' to 'Mutating variables'
- added section on dates and date time in python and pandas


_episodes/01-starting-with-data.md
```diff
@@ -31,8 +31,10 @@ and they can replicate the same analysis.
 
 To help the lesson run smoothly, let's ensure everyone is in the same directory.
 This should help us avoid path and file name issues. At this time please
-navigate to the workshop directory. If you working in IPython Notebook be sure
-that you start your notebook in the workshop directory.
+navigate to the workshop directory. 
+If working in Spyder open the project you created in the previous session. You can do this under the _Projects_ tab.
```
stijnvanhoey: As Spyder (althought as part of the full Anaconda install included) is not part of the [setup](http://www.datacarpentry.org/python-ecology-lesson/setup/) and is 'just another IDE', maybe the information about Spyder can confuse new learners?

katrintirok: thanks, I agree, the setup section would need to be updated as well. I think the before-we-start session could also be written more general - setting up a working directory and finding help is general, whether using an IDE or notebook. Could then rather introduce spyder (as an example for an IDE and part of Anaconda) in setup, where also notebooks are introduced. Or rather move the notebook intro to "Before-we-start" episode? I am happy to rework this material, some more inputs from discussion would be good.

stijnvanhoey: Introducing both could be a bit overwhelming to learners, but I do understand that instructors have a preference to teach in either one or the other. 
Having all the setup/environment in the `before-we-start` with two parallel options seems a legit option. I think it is important that all other episodes are *independent* from the environment, as this could cause confusion.

wrightaprilm: The established tool for the Carpentries is the Jupyter Notebook. I think the way to handle this might be to put  text explaining the Jupyter Notebook in the "Before we start". I'm happy to contribute this text - there is a lesson set on using Jupyter for reproducible research [in development]( https://reproducible-science-curriculum.github.io/workshop-RR-Jupyter/#iii-introduction-to-the-jupyter-notebook), so having a standard set of language for referring to this tool is likely to be both helpful and useful. Then, the text about Spyder can go in the Instructor Guide. So it's there if you need it; instructors can easily paste it in when they fork the lesson for their workshop. Would this be workable?

katrintirok: Hi, thanks for the feedback. I will go through the lesson again and take out any remarks to spyder. Nice idea to add some info on spyder in the instructor guide. Currently, there is info on how to use the Jupyter notebook in the setup session. We could leave it in there and make the 'Before we start' episode (and all other episodes) independent from the environment. I revised the 'Before we start' episode (#261) last week to make it more general, just saw that you merged it already, great. It is on my to do list to also go through all episodes to check that they line up ... 
I like the Jupyter notebook introduction from the reproducible research lesson.

trallard: I was just having a look at this 👀 I am completely in favour of having the Jupyter notebook instructions only in this section. And if absolutely necessary have the Spyder remarks in the instructor notes although I think most folks prefer the use of the notebooks for these materials. Anyway @katrintirok let me know if you need any help adding these bits

katrintirok: Just removed all comments in regard to spyder. 
The existing conflict arises from the changed variable 'plot_id'  to 'site_id' which introduced some inconsistencies that should be solved first (see issue #279)

maxim-belkin: Current conflict is:

```
<<<<<<< update-starting-with-data
```python
surveys_df.head() # The head() function displays the first several lines of a file. It is discussed below.
   
   record_id  month  day  year  plot_id species_id sex  hindfoot_length  \
=======
>>> surveys_df.head() # The head() function displays the first several lines of a file. It
		      # is discussed below.
   record_id  month  day  year  site_id species_id sex  hindfoot_length  \
>>>>>>> gh-pages
```

maxim-belkin:
@katrintirok, please split this PR into smaller ones so we can easily discuss changes quicker.
Let me know if you need any help with this procedure.

PR #261, opened 2018-04-09
Adding Before-we-start episode

katrintirok:
adding episode Before-we-start for general background on what python is, how to set up environment, where to find help etc. Split from #257


ErinBecker:
I like this. I've found the corresponding episode in the R Ecology materials to be very helpful to learners when teaching those materials. I haven't taught the Python Ecology lesson, so I can't say for sure that this would be helpful there, but it seems to me like it would be.

wrightaprilm:
I like this a lot, as well. But it does introduce a re-numbering issue for the lessons, so I think what I'm going to do is merge in the changes to the plotting lesson (making the corrections noted in the testing issue), then merge this, then address changes to numbering, then finally take a look at the intro to python lesson.

wrightaprilm:
Thanks so much for this, @katrintirok. Merged. It looks like it actually doesn't cause a conflict with the numbering, so I'll just leave it as 00 for now.

PR #264, opened 2018-04-14
Plotnine fix

stijnvanhoey:
After the merge of PR #248, I noticed that a section of the faceting section was not rendering well:
image
Actually, this was due to a set of old text/paragraphs still in the new text, whereas the actual challenge was already reformulated just below that section. This PR removes the redundant text from the old text (and fixes an additional small typo).


wrightaprilm:
Ah, good, thanks. I resolved a conflict via the web interface, which I don't normally do. I probably goofed something in the process.

stijnvanhoey:
No problem and thanks for the quick response.

PR #265, opened 2018-04-16
Change location plots to sites for workshop sample data to avoid ambiguity with data vis plotting

EdMBarry:
---
Part of instructor training checkout. Tried as best as I could to judge whether plots corresponded to site or data vis. Could not change axes labels in the figs directory. Renamed \*plot\*.png files  but could not find their source files, so this could create some headaches.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -82,7 +82,7 @@ guidelines:
   exporting high-quality plots.
 - [ ] Lesson 7 repeats some matplot content, but with a new dataset. Rework to
   use the previous lesson data.
-- [ ] The column name `plot_id` in surveys.csv confused a few participants who
+- [x] The column name `plot_id` in surveys.csv confused a few participants who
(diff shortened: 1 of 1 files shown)


katrintirok:
just was working on 01-intro-do-data episode and tumbled over the 'plot --> site' update. I understand there was an issue that some participants relating plot to a graph and not land. However, in the data set on figshare the variable is still called 'plot_id', and at the beginning of the intro-to-data episode surveys.csv is supposed to be downloaded from figshare. In the setup episode, the files obviously link to what is on the GitHub repository, where the variable is now 'site_id'. 

We should also remember that the same data are used in the other lessons of the ecology stream, i.e. data organisation in spreadsheets, OpenRefine, R and SQL.

I suggest to go back to 'plot_id' to keep everything consistent.

One could suggest to make a change to the portal teaching data base on figshare if the variable naming causes serious problems.

PR #271, opened 2018-04-24
P9 images

tmorrell:
Adding images back to plotnine version of plotting lesson.  Also made more of the plots in color.


tmorrell:
Rendered version available at https://tmorrell.github.io/python-ecology-lesson/06-visualization-ggplot-python/

PR #283, opened 2018-05-17
change site_id back to plot_id in self-hosted files

wrightaprilm:
Self-explanatory. If you could review @maxim-belkin, it would be much appreciated.


astrochun:
Thanks. That's better, though perhaps we should the Challenge -Statistics content:
1. Create a list of unique site ID’s found in the surveys data. Call it site_names.

I think learners will be confused.

maxim-belkin:
Can you clarify this:
> That's better, though perhaps we should the Challenge -Statistics content:

astrochun:
@maxim-belkin: I've gone ahead and done a pull request for a suggested fix #284. I'm referring to one of the challenge questions, which says:

Challenge - Statistics
Create a list of unique site ID’s found in the surveys data. Call it site_names. How many unique sites are there in the data? How many unique species are in the data?

- While you do provide a description for plot_id in the table, it's easy for learners to miss this.  Someone who looks at the csv file or the DataFrame will try to look for something connected to "site ID," but they could then be confused.

maxim-belkin:
I used imagemagick to process images:
```
convert right-join.png -flatten -trim +repage -fuzz 10% -transparent white -resize "600>" right-join-trimmed.png
```

PR #286, opened 2018-05-20
Update-question-repeat-epsiodes00.md

erikamias:
Line 4: change to 
- "What are the benefits of programming?"
- "What are the foundational elements of coding in Python?"
(Questions are duplicated from episode 00-Before-we-start/ , Also numbering of lessons is correct in schedule but numbering in episodes is out of sequence)

### DIFF
--- a/_episodes/00-short-introduction-to-Python.md
+++ b/_episodes/00-short-introduction-to-Python.md
@@ -14,19 +14,6 @@ keypoints:
     - "FIXME"
 ---
 
-# The Basics of Python
-
-Python is a general purpose programming language that supports rapid development
-of scripts and applications.
-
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Erika, the questions that we ask in each episode have to be directly related to the the specific episode and be very limited in scope.

- "What are the benefits of programming?" is a very general question and we are not, actually, trying to answer it here.
- "What are the foundational elements of coding in Python?" is a very broad question and, again, this episode is not covering that. In fact "foundational elements of coding in Python" is the learning goal of the entire lesson.

Therefore, I'm closing this PR without merging it, but thank you for your contribution anyway!

PR #288, opened 2018-05-22
removed duplicated text

erikamias:
"Basics of Python" text duplicated from "Before we start"


maxim-belkin:
Thanks, @erikamias !

PR #301, opened 2018-06-19
Changed format for Jupyter/Spyder headers. Instructor-training assignment

plestran:


maxim-belkin:
Thanks @plestran! I think this change makes sense, but it would be nice to include some justification in the commit message. Can you write it up?

plestran:
Good call! I added an additional comment to my commit.

PR #302, opened 2018-06-19
Fix rendering of some pages

shwina:
Right now these pages do not render correctly:

http://www.datacarpentry.org/python-ecology-lesson/reference/
http://www.datacarpentry.org/python-ecology-lesson/setup/
http://www.datacarpentry.org/python-ecology-lesson/conduct/

Removing "permalink" from the header seems to work. I also don't see permalink being specified in the lesson-example:

https://github.com/carpentries/lesson-example

### DIFF
--- a/CONDUCT.md
+++ b/CONDUCT.md
@@ -1,8 +1,8 @@
 ---
 layout: page
 title: "Contributor Code of Conduct"
-permalink: /conduct/
 ---
+
 As contributors and maintainers of this project,
 we pledge to follow the [Carpentry Code of Conduct][coc].
 
--- a/reference.md
+++ b/reference.md
@@ -1,6 +1,5 @@
 ---
 layout: reference
-permalink: /reference/
 ---
 
 ## Glossary
--- a/setup.md
+++ b/setup.md
@@ -1,7 +1,6 @@
 ---
 layout: page
 title: Setup
-permalink: /setup/
 ---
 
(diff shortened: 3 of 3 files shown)


maxim-belkin:
Cool! Thanks @shwina. I think the issue is caused by the change in `root` value, but your fix works.

maxim-belkin:
What I've tested is that when I run `make serve` locally, generated links in the site are correct.
My solution would've been to add `root ..` instead of deleting permalink.

fmichonneau:
I'm going to push an update to the template across the repositories soon, and this should get fixed in the process. 

My guess is that it's related to an out of date _config.yml compared to what is expected by the YAML headers in the files.

shwina:
OK. I wasn't sure about root as that isn't used in the lesson example. In fact it looks like it was explicitly removed recently:

ab1ea985a878639624685dec8ce13d1b33a759cc

maxim-belkin:
> My guess is that it's related to an out of date _config.yml compared to what is expected by the YAML headers in the files.

I merged `_config.yml` not so long ago, so it shouldn't be that.

> ab1ea985a878639624685dec8ce13d1b33a759cc

which repo is that?

shwina:
Sorry:
https://github.com/carpentries/lesson-example/commit/ab1ea985a878639624685dec8ce13d1b33a759cc

Issue #306, opened 2018-07-02
Incompatible file (SQLite) among different download links


kokbent:
This is regarding the download links for data files in the setup page: http://www.datacarpentry.org/python-ecology-lesson/setup.html

The portal_mammals.sqlite file from the figshare (the teaching database), is not the same as the one you get from the following download links that are provided in the same section:
1. http://www.datacarpentry.org/python-ecology-lesson/data/portal_mammals.sqlite
2. https://minhaskamal.github.io/DownGit/#/home?url=https://github.com/datacarpentry/python-ecology-lesson/tree/gh-pages/data (This is the one associated with "Download these files to your computer either by clicking this link").

In particular, the SQLite file that comes with the two links do not have "hindfoot_length" column in "surveys" table. And instead of "weight", "plot_id" and "species_id" columns, it has "wgt", "plot" and "species". It caused some confusion as we transition from Python to SQL module in a workshop last week.

I recommend to standardize all data files among all modules to the figshare version, and the two links should be removed from the page. From what I understand, the different versions of SQLite file do not affect the "Accessing SQLite Databases Using Python and Pandas" section.

maxim-belkin:
@kokbent, I'm not sure what caused this discrepancy but in order for the lesson to be complete, it has to provide data files. The two "links" that you specify point to the same data -- just two different ways to get it.
In the notes to the data on `figshare` page, it says that the data is, actually, stored on GitHub at https://github.com/weecology/portal-teachingdb

So, here is what we can do:

1. Add an additional minhashamal's link to that repo:
https://minhaskamal.github.io/DownGit/#/home?url=https://github.com/weecology/portal-teachingdb
2. Add a link to download that repository as zip archive: https://github.com/weecology/portal-teachingdb/archive/master.zip
2. Update our files in `data` directory 
    - Update instructions where necessary 

CC @wrightapril @trallard

wrightaprilm:
Thanks for your patience on this; I've been traveling. I'm somewhat confused by the issue here. As @maxim-belkin notes, these two links point to the same data. A diff of the two shows them to be the same, and it looks like they have the same columns. Could you link to the workshop webpage? I'd like to reproduce the error myself.

kokbent:
Here's the "Data" section of the setup page: https://datacarpentry.org/python-ecology-lesson/setup.html
> Data for this lesson is from the Portal Project Teaching Database - available on FigShare.
[...]

PR #309, opened 2018-08-07
move to figshare links

wrightaprilm:
Fix issue #306. Pretty simple, but it would be great to have a second set of eyes on.

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -7,18 +7,18 @@ title: Setup
 > Data for this lesson is from the Portal Project Teaching Database -
 > [available on FigShare](https://figshare.com/articles/Portal_Project_Teaching_Database/1314459).
 >
-> We will use the eight files listed below for the data in this lesson.  Download these files to your computer either by clicking [this link ](https://minhaskamal.github.io/DownGit/#/home?url=https://github.com/datacarpentry/python-ecology-lesson/tree/gh-pages/data), which will give you everything in a single compressed file.  You'll need to unzip this file after downloading it.
+> We will use the eight files listed below for the data in this lesson.  Download these files to your computer either by clicking [this link ](https://minhaskamal.github.io/DownGit/#/home?url=https:%2F%2Fgithub.com%2Fweecology%2Fportal-teachingdb), which will give you everything in a single compressed file.  You'll need to unzip this file after downloading it.
 >
 > Or download each file indvidually with the following links:
 >
-> - [surveys.csv]({{ page.root }}/data/surveys.csv)
-> - [species.csv]({{ page.root }}/data/species.csv)
+> - [surveys.csv](https://ndownloader.figshare.com/files/10717177)
+> - [species.csv](https://ndownloader.figshare.com/files/3299483)
 > - [speciesSubset.csv]({{ page.root }}/data/speciesSubset.csv)
 > - [surveys2001.csv]({{ page.root }}/data/survey2001.csv)
 > - [surveys2002.csv]({{ page.root }}/data/survey2002.csv)
-> - [plots.csv]({{ page.root }}/data/plots.csv)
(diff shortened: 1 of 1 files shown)


setup.md
```diff
@@ -7,18 +7,18 @@ title: Setup
...
+> - [surveys.csv](https://ndownloader.figshare.com/files/10717177)
+> - [species.csv](https://ndownloader.figshare.com/files/3299483)
 > - [speciesSubset.csv]({{ page.root }}/data/speciesSubset.csv)
 > - [surveys2001.csv]({{ page.root }}/data/survey2001.csv)
 > - [surveys2002.csv]({{ page.root }}/data/survey2002.csv)
-> - [plots.csv]({{ page.root }}/data/plots.csv)
+> - [plots.csv](https://ndownloader.figshare.com/files/3299474)
 > - [bouldercreek_09_2013.txt]({{ page.root }}/data/bouldercreek_09_2013.txt)
-> - [SQL Database]({{ page.root }}/data/portal_mammals.sqlite)
+> - [SQL Database](https://ndownloader.figshare.com/files/11188550)
```
maxim-belkin: @wrightaprilm, do we want to keep copies of the data files in this repo (in the `data` folder)?

maxim-belkin:
Adding bullet points from #306 here

- [ ] Run through the lessons quickly and make sure they work with the Figshare data before merging
- [ ] Move surveys2001 and 2002 to sample output, or just delete them
- [ ] Add the other two datafiles to a figshare project so we have an external. Alternatively, I think it would be OK to delete speciesSubset, and have them take the first couple rows of the species.csv file as their dataset for the join activity. It's really just provided so there is a small enough dataset to see the different joins clearly and non-overwhelmingly.

setup.md
```diff
@@ -7,18 +7,18 @@ title: Setup
 > Data for this lesson is from the Portal Project Teaching Database -
 > [available on FigShare](https://figshare.com/articles/Portal_Project_Teaching_Database/1314459).
 >
-> We will use the eight files listed below for the data in this lesson.  Download these files to your computer either by clicking [this link ](https://minhaskamal.github.io/DownGit/#/home?url=https://github.com/datacarpentry/python-ecology-lesson/tree/gh-pages/data), which will give you everything in a single compressed file.  You'll need to unzip this file after downloading it.
+> We will use the eight files listed below for the data in this lesson.  Download these files to your computer either by clicking [this link ](https://minhaskamal.github.io/DownGit/#/home?url=https:%2F%2Fgithub.com%2Fweecology%2Fportal-teachingdb), which will give you everything in a single compressed file.  You'll need to unzip this file after downloading it.
```
maxim-belkin: no need to change `/` to `%2F`

_episodes/09-working-with-sql.md
```diff
@@ -66,7 +66,7 @@ con = sqlite3.connect("data/portal_mammals.sqlite")
 cur = con.cursor()
 
 # Return all results of query
-SELECT plot_id FROM plots WHERE plot_type="Control"')
+cur.execute('SELECT plot_id FROM plots WHERE plot_type="Control"')
```
maxim-belkin: It might be better to do it in a separate PR

wrightaprilm: Just rolled it back.

wrightaprilm:
OK, I ran through everything. It all works with the ndownloader links, so I think we ought to merge this. Regarding points 2 and 3, I think we should just delete the surveys files. We can also delete plots.csv - it's no longer used. I agree with removing speciesSubset and just having them make it.

maxim-belkin:
OK. Let's merge this PR and create a PR that deletes the files you mention.

wrightaprilm:
Sounds like a plan.

PR #317, opened 2018-09-12
Setup

wrightaprilm:
Remove extraneous files.


maxim-belkin:
@wrightaprilm I wrapped long lines and replaced `%2F` with `/` in the link.

wrightaprilm:
Good call!

PR #318, opened 2018-09-21
Episode 8: add time estimates

wrightaprilm:
I'm not sure why the spacings are different. Does that look OK to you, @maxim-belkin?


maxim-belkin:
Your text editor removed empty whilespaces at  the end of lines (which is a good thing).

wrightaprilm:
All right, if you approve, I'll merge it.

PR #325, opened 2018-11-11
small grammatical change

raynamharris:
this is a small grammatical change as suggested during the translation to Spanish in https://github.com/Carpentries-ES/python-ecology-lesson-es/pull/30#discussion_r232508766

An alternative could be to say "The SciPy ecosystem..."

### DIFF
--- a/_episodes/00-before-we-start.md
+++ b/_episodes/00-before-we-start.md
@@ -11,7 +11,7 @@ objectives:
   - "Demonstrate how to provide sufficient information for troubleshooting with the Python user community."
 keypoints:
   - "Python is an open source and platform independent programming language"
-  - "SciPy ecosystem for Python provides the tools necessary for scientific computing"
+  - "SciPy is an ecosystem for Python provides the tools necessary for scientific computing"
   - "Jupyter Notebook and the Spyder IDE are great tools to code in and interact with Python with its large community it is easy to find help in the internet"
 ---
 


orchid00:
I think it sounds better with which:

SciPy is an ecosystem for Python which provides the tools necessary for scientific computing.

maxim-belkin:
What is the motivation for this change?

raynamharris:
One our our Spanish translators thought it was missing a word to make a complete sentence. But, the more I read it, I think it's okay as is. I think it's okay if you want to close without merging @maxim-belkin

maxim-belkin:
We're conveying two messages here: 1. that SciPy is an ecosystem (not just one "thing"). 2. that it provides tools for scientific computing. Then, I don't think that (1) deserves a separate keypoint on its own as this is a rather general statement. So, we remove emphasis on the fact that SciPy **is** an ecosystem (by not saying "SciPy **is** an ecosystem) and state it as if this is a known fact.

Nonetheless, I think the right fix is the one you mention in your original message, that is: `The SciPy ecosystem`. Then, the ecosystem is not really `for Python` (it's within it), so I think the right change should really be:
```
The SciPy ecosystem provides tools for scientific computing in Python.
```

This is more or less the same as what is said on their webpage: https://www.scipy.org/about.html

However, this statement is rather weak. I think we could say:
```
The SciPy ecosystem provides tools that enable scientific computing in Python.
```
but this is open for a discussion.

raynamharris:
Nice. I really like your last option "The SciPy ecosystem provides tools that enable scientific computing in Python."

_episodes/00-before-we-start.md
```diff
@@ -11,7 +11,7 @@ objectives:
   - "Demonstrate how to provide sufficient information for troubleshooting with the Python user community."
 keypoints:
   - "Python is an open source and platform independent programming language"
-  - "SciPy ecosystem for Python provides the tools necessary for scientific computing"
+  - "SciPy is an ecosystem for Python provides the tools necessary for scientific computing"
```
wrightaprilm:
```suggestion
  - "The SciPy ecosystem provides tools that enable scientific computing in Python."
```

wrightaprilm:
I went ahead and inserted that language as a suggestion. Feel free to accept if it looks good, @raynamharris. If it still looks off, we can discuss more.

wrightaprilm:
Hey @raynamharris. I'm just getting back into the swing of things over here - it looks like there's just one outstanding issue to resolve. I made a suggestion; if it looks good, go ahead and accept and I'll merge this.

tobyhodges:
This suggested change was rendered obsolete when #415 was merged.

PR #330, opened 2018-11-20
fix typo episode 03

orchid00:
referencing https://github.com/Carpentries-ES/python-ecology-lesson-es/issues/8#issuecomment-440413436 thanks to @welozano


wrightaprilm:
Looks good! I'll merge now.

I agree on using the personal fork for pull requests. I was a bit lax on this for a long time, since I was maintaining more or less on my own. But now, with all the activity, I think it will keep the repo smaller (and easier for folks to clone) and make a clearer record of the context in which to understand changes. Thanks!

PR #334, opened 2018-11-21
deleted outdated reference

NPalopoli:
Deleted specific mention to 'Mac OS with Homebrew' as installation in [Setup](https://datacarpentry.org/python-ecology-lesson/setup.html) is now based on conda/Anaconda for all platforms.


maxim-belkin:
Thanks again and again, @NPalopoli!

PR #335, opened 2018-11-21
guide.md: update Q&A in "Dictionaries Challenges" section

NPalopoli:
Updated question and answer in section **Dictionaries Challenges** based on corresponding section in episode [01-short-introduction-to-Python.md](https://github.com/datacarpentry/python-ecology-lesson/blob/gh-pages/_episodes/01-short-introduction-to-Python.md)

### DIFF
--- a/_extras/guide.md
+++ b/_extras/guide.md
@@ -41,9 +41,9 @@ installation. Otherwise, it will tell you that the system is good to go and read
 
 ### Dictionaries Challenges
 
-* Changing dictionaries: 2. Reassign the second value.
+* Changing dictionaries: 2. Reassign the value that corresponds to the key `2`.
 
-Make sure it is also clear that access to 'the second value' is actually just about the key name. Add for example `rev[10] = "ten"` to clarify it is not about the position.
+Make sure it is also clear that access to 'the value that corresponds to the key `2`' is actually just about the key name. Add for example `rev[10] = "ten"` to clarify it is not about the position.
 
 ~~~
 rev
--- a/_extras/guide.md
+++ b/_extras/guide.md
@@ -40,12 +40,19 @@ installation. Otherwise, it will tell you that the system is good to go and read
 	`tuple`
 
 ### Dictionaries Challenges
-* Can you do reassignment in a dictionary? Give it a try.
+
+* Changing dictionaries: 2. Reassign the second value.
 
 Make sure it is also clear that access to 'the second value' is actually just about the key name. Add for example `rev[10] = "ten"` to clarify it is not about the position.
 
-* You should see the following output:
-`{1: 'one', 2: 'two', 3: 'three'}`
+~~~
+rev
+~~~
+{: .language-python}
+~~~
+{1: 'one', 2: 'two', 3: 'three'}
+~~~
+{: .output}
 
 ~~~
 rev[2] = "apple-sauce"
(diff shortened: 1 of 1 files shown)


_extras/guide.md
```diff
@@ -40,12 +40,19 @@ installation. Otherwise, it will tell you that the system is good to go and read
 	`tuple`
 
 ### Dictionaries Challenges
-* Can you do reassignment in a dictionary? Give it a try.
+
+* Changing dictionaries: 2. Reassign the second value.
```
maxim-belkin: 1. What does `2.` mean?
2. `Second value` does not sound right as dictionary is not a sequence of values but of key-value pairs.

wrightaprilm: I agree with this "Reassign the value of the second key" might be what we're looking for here.

maxim-belkin: Thank you, April. However, we can not even use "the second key" as dictionaries are not ordered. We have to be very specific with dictionaries and say `value that corresponds to key of <blah>` (blah = 2 in this case).

wrightaprilm: Good catch - easy to forget. I like

"Reassign the value that corresponds to the key "2"."

NPalopoli:
Nice one! Thanks Maxim and April. I've created #336 to address this in the
lesson.

--
Nicolás Palopoli
Structural Bioinformatics Group
Universidad Nacional de Quilmes - CONICET
http://orcid.org/0000-0001-7925-6436

_extras/guide.md
maxim-belkin: Please don't hate my but... no double quotes around 2 because `2` and `"2"` are two different and legitimate keys.

NPalopoli:
I agree, thanks! And I think we can avoid the double quotes without
allowing for confusions.

I'm not sure if the previous changes have been accepted or not yet. Please
go ahead with the PR for removing the double quotes and close the issue if
done.

_extras/guide.md
wrightaprilm: Resolved by #336

wrightaprilm: My mistake, this aspect is not resolved by #336. Could we change this  to 

" Reassign the value that corresponds to the key `2`."

NPalopoli: Thank you @wrightaprilm and @maxim-belkin. Should be correct now.

wrightaprilm:
Great, thanks for fixing that. I'm happy if you're happy, @maxim-belkin.

_extras/guide.md
```diff
@@ -40,12 +40,19 @@ installation. Otherwise, it will tell you that the system is good to go and read
 	`tuple`
 
 ### Dictionaries Challenges
-* Can you do reassignment in a dictionary? Give it a try.
 
-Make sure it is also clear that access to 'the second value' is actually just about the key name. Add for example `rev[10] = "ten"` to clarify it is not about the position.
+* Changing dictionaries: 2. Reassign the value that corresponds to the key `2`.
```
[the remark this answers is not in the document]
wrightaprilm: There's a one in the lesson, but no one has apparently felt moved to add it to the guide (it's just printing the dictionary).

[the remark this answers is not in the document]
maxim-belkin: OK, I see its purpose now. Let's merge this as is then. I'll submit my suggestions (at some point)

wrightaprilm:
Thanks @NPalopoli!

PR #338, opened 2018-11-22
03-index-slice-subset.md: use proper dataframe name

NPalopoli:
Replaced generic dataframe name `df` by lesson-specific `surveys_df`


wrightaprilm:
I'm happy with this if you are, @maxim-belkin

PR #339, opened 2018-11-22
minor: double spaces to separate

NPalopoli:
Finished lines with double spaces in Selection Challenges, so they display in different lines


wrightaprilm:
Thanks!

PR #340, opened 2018-11-22
03-index-slice-subset.md: improve wording

welozano:
The  sentence refers to x > 5, I think there is an extra "not"

Wil

### DIFF
--- a/_episodes/03-index-slice-subset.md
+++ b/_episodes/03-index-slice-subset.md
@@ -458,9 +458,9 @@ x == 5
 ~~~
 {: .language-python}
 
-When we ask Python what the value of `x > 5` is, we get `False`. This is
-because the condition, `x` is greater than 5, is not met since `x` is equal
-to 5.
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Thanks, @welozano! good catch! We can merge this as is. However, I'd like to suggest that we spend a bit more time and rewrite this paragraph a bit as it's not super well written.

My suggestion (feel free to improve):

> When we ask Python whether `x` is greater than `5`, it returns `False`. This is Python's way to say "No". Indeed, the value of `x` is `5`, and `5` is not greater than `5`.

@wrightaprilm, please chime in. :)

wrightaprilm:
I like your replacement text, Maxim. If you're willing to add it in, @welozano, we would very much appreciate it!

welozano:
Hi, the suggestion looks pretty good. I just added as @maxim-belkin suggested.
Wil

wrightaprilm:
Great, thanks @welozano!

PR #342, opened 2018-11-23
guide.md: added consistency to 04-data-types-and-format section

NPalopoli:
Added Challenge guidelines (taken from the lesson) and language markup to section 04-data-types-and-format for consistency with the other sections.

### DIFF
--- a/_extras/guide.md
+++ b/_extras/guide.md
@@ -258,7 +258,7 @@ surveys_df.isnull()
 ~~~
 {: .language-python}
 
-If the students have trouble generating the output, or anything happens with that, there is a file called "sample output" that contains the data file they should generate.
+If the students have trouble generating the output, or anything happens with that, the folder `sample_output` in this repository contains the file `surveys_complete.csv` with the data they should generate.
 
 ## 05-merging-data
 
--- a/_extras/guide.md
+++ b/_extras/guide.md
@@ -251,7 +251,7 @@ Pandas cannot convert types from float to int if the column contains NaN values.
 
 ### Challenge - Counting
 
-* Count the number of missing values per column. Hint: The method .count() gives you the number of non-NA observations per column. Try looking to the .isnull() method.
+* Count the number of missing values per column. Hint: The method `.count()` gives you the number of non-NA observations per column. Try looking to the `.isnull()` method.
 
 ~~~
 surveys_df.isnull()
--- a/_extras/guide.md
+++ b/_extras/guide.md
@@ -245,7 +245,9 @@ stack_selection.columns = stack_selection.columns.droplevel()
 
 ### Challenge - Changing Types
 
-* Try converting the column `plot_id` to floats using `surveys_df.plot_id.astype("float")`. Then, try converting the contents of the `weight` column to an integer type. What error message does Pandas give you? What does this error mean?
+* Try converting the column `plot_id` to floats using `surveys_df.plot_id.astype("float")`. 
+Then, try converting the contents of the `weight` column to an integer type. 
+What error messages does Pandas give you? What do these errors mean?
 
 Pandas cannot convert types from float to int if the column contains NaN values.
 
--- a/_extras/guide.md
+++ b/_extras/guide.md
@@ -245,7 +245,7 @@ stack_selection.columns = stack_selection.columns.droplevel()
 
 ### Challenge - Changing Types
 
-* Try converting the column `plot_id` to floats using `surveys_df.plot_id.astype("float")`. Next try converting `weight` to an integer. What goes wrong here? What is Pandas telling you?
+* Try converting the column `plot_id` to floats using `surveys_df.plot_id.astype("float")`. Then, try converting the contents of the `weight` column to an integer type. What error message does Pandas give you? What does this error mean?
 
 Pandas cannot convert types from float to int if the column contains NaN values.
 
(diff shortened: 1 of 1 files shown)


_extras/guide.md
```diff
@@ -245,13 +245,20 @@ stack_selection.columns = stack_selection.columns.droplevel()
 
 ### Challenge - Changing Types
 
+* Try converting the column `plot_id` to floats using `surveys_df.plot_id.astype("float")`. Next try converting `weight` to an integer. What goes wrong here? What is Pandas telling you?
+
 Pandas cannot convert types from float to int if the column contains NaN values.
 
 ### Challenge - Counting
+
+* Count the number of missing values per column. Hint: The method .count() gives you the number of non-NA observations per column. Try looking to the .isnull() method.
```
maxim-belkin: use backticks around .count() and .isnull() so that they appear as `.count()` and `.isnull()`

_extras/guide.md
```diff
@@ -245,13 +245,20 @@ stack_selection.columns = stack_selection.columns.droplevel()
 
 ### Challenge - Changing Types
 
+* Try converting the column `plot_id` to floats using `surveys_df.plot_id.astype("float")`. Next try converting `weight` to an integer. What goes wrong here? What is Pandas telling you?
```
maxim-belkin: Next -> then

maxim-belkin: `Then try converting `weight` column to an integer type`

maxim-belkin: I'd remove `What goes wrong here?`... and say `What does _pandas_ tell you when you try doing that?` but I'll leave it to @wrightaprilm to decide what is the best wording here

wrightaprilm: How about "What error message does Pandas give you? What does this error mean?" Emphasizing specifically that we want them to read and comprehend the output, which in this case is very helpful and informative!

_extras/guide.md
wrightaprilm:
```suggestion
* Count the number of missing values per column. Hint: The method `.count()` gives you the number of non-NA observations per column. Try looking to the `.isnull()` method.
```

If you just want to accept this change, @NPalopoli, it contains the correct change.

_extras/guide.md
```diff
@@ -245,13 +245,20 @@ stack_selection.columns = stack_selection.columns.droplevel()
...
+* Count the number of missing values per column. Hint: The method .count() gives you the number of non-NA observations per column. Try looking to the .isnull() method.
+
+~~~
 surveys_df.isnull()
+~~~
+{: .language-python}
 
-If the students have trouble generating the output, or anything happens with that, there is a file
-called "sample output" that contains the data file they should generate.
+If the students have trouble generating the output, or anything happens with that, there is a file called "sample output" that contains the data file they should generate.
```
wrightaprilm: "there is a folder called `sample_output`, which contains the data file they should generate. It is not included with the data in the setup steps."

Most instructors don't have the learners clone the repo, as Git and GitHub are not part of the Data Carpentry core. So learners typically don't have this file, but the instructor could provide it.

NPalopoli: Nice call April. Perhaps this is a good wording?

```suggestion
If the students have trouble generating the output, or anything happens with that, there is a file called "sample output" that contains the data file they should generate. It is not included with the data in the setup steps, so remember to have it at hand and provide it if needed.
```

_extras/guide.md
NPalopoli: Thanks @maxim-belkin  and @wrightaprilm. Is this a good wording to commit?

```suggestion
* Try converting the column `plot_id` to floats using `surveys_df.plot_id.astype("float")`. Then, try converting the contents of the `weight` column to an integer type. What error message does Pandas give you? What does this error mean?
```

_extras/guide.md
maxim-belkin: @NPalopoli, note "folder" vs "file".
Another note: "there is" does not answer the question of "where". I suggest saying something like `This repository has a folder called...`

maxim-belkin:
Please change all `Try doing something` to `Do something`, for example: `Try converting` to `Convert`.

_extras/guide.md
wrightaprilm: I like this wording!

wrightaprilm:
I actually like the `try` language, particularly for things where we are asking for an action that we know will produce an error.

maxim-belkin:
Ha! Agreed! Was not realizing we're talking about stuff that is going to fail. Please leave those `Try`.

_extras/guide.md
NPalopoli: Thank you @maxim-belkin once and again! Should be correct now.

wrightaprilm:
These changes satisfy my concerns. I'm happy if you're happy, @maxim-belkin.

_extras/guide.md
```diff
@@ -245,13 +245,20 @@ stack_selection.columns = stack_selection.columns.droplevel()
 
 ### Challenge - Changing Types
 
+* Try converting the column `plot_id` to floats using `surveys_df.plot_id.astype("float")`. Then, try converting the contents of the `weight` column to an integer type. What error message does Pandas give you? What does this error mean?
```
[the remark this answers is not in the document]
NPalopoli: Already committed those changes. Thank you for pointing out @maxim-belkin!

wrightaprilm:
Great, thanks everyone!

PR #343, opened 2018-11-23
05-merging-data.md: use proper file name

NPalopoli:
Please delete the text below before submitting your contribution.


maxim-belkin:
I believe the filename is correct. Please clarify why you think it is not.

wrightaprilm:
I actually like this. As-is in the lesson, we say a plots CSV file. But not that it's actually called `plots.csv`. I'd feel good about merging this, since it's more explicit.

maxim-belkin:
Thank you, @NPalopoli!

PR #344, opened 2018-11-23
guide.md: removing deprecated challenge

NPalopoli:
Removed mention and response to challenge that was removed from lesson 06-loops-and-functions.md

PR #347, opened 2018-11-24
extra_challenges.md: new file

NPalopoli:
Following @maxim-belkin suggestion in #334, this file would store all extra challenges (exercises) that may be useful for instructors in the future.

### DIFF
--- a/_extras/extra_challenges.md
+++ b/_extras/extra_challenges.md
@@ -8,12 +8,31 @@ permalink: /extra_challenges/
 
 A collection of challenges that have been either removed from or not (yet) added to the main lesson. 
 
-## 06-loops-and-functions
-
-### Looping Over Dataframe Challenges
-
-* What happens if there is no data for a year in the sequence (for example, imagine we had used 1976 as the start year in `range`)
-~~~
-An empty file with only the headers
-~~~
-{: .output}
+> ## Looping Over Dataframe
+>
+> (Please refer to lesson `06-loops-and-functions.md`)
+>
+> The file `surveys.csv` in the `data` folder contains 25 years of data from surveys,
+> starting from 1977. We can load the data and print all the years surveyed using a `for` loop:
+> 
+> ~~~
+> import pandas as pd
+> 
+> # Load the data into a DataFrame
+> surveys_df = pd.read_csv('data/surveys.csv')
+> 
+> # Loop through a sequence of years and print the year
+> start_year = 1977
+> end_year = 2002
+> for year in range(start_year, end_year+1):
+>     print(str(year))
+> ~~~
+> {: .language-python}
+>
+> What happens if there is no data for a year in a sequence? For example,
+> imagine we used `1976` as the start year in `range`
+>
+> > ## Solution
+> > An empty file with only the headers  
+> {: .solution}  
+{: .challenge}
--- a/_extras/extra_challenges.md
+++ b/_extras/extra_challenges.md
@@ -30,7 +30,7 @@ A collection of challenges that have been either removed from or not (yet) added
 > {: .language-python}
 >
 > What happens if there is no data for a year in a sequence? For example,
-> imagine we used `1976` as the start year in `range`
+> imagine we used `1976` as the year in `surveys_df[surveys_df.year == year]`
 >
 > > ## Solution
 > > An empty file with only the headers  
--- a/_extras/extra_challenges.md
+++ b/_extras/extra_challenges.md
@@ -8,12 +8,13 @@ permalink: /extra_challenges/
 
 A collection of challenges that have been either removed from or not (yet) added to the main lesson. 
 
-> ## Looping Over Dataframe
+> ## Looping Over DataFrame
 >
 > (Please refer to lesson `06-loops-and-functions.md`)
 >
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Thanks, @NPalopoli. Now it is a good time to familiarize yourself with Carpentries' format for challenges:

```
> ## Title (use two pound signs here)
> 
> Question to learners
>
> > ## Solution
> > this is the answer
> > ~~~
> > some + python = code
> > ~~~
> > {: .language-python}
> {: .solution}
{: .challenge}
```

maxim-belkin:
Ping, @NPalopoli!
Here is how you can convert the challenge in this PR into the correct format:

```markdown
> ## Looping Over Dataframe
>
> What happens if there is no data for a year in a sequence? For example,
> imagine we used `1976` as the start year in `range`
>
> > ## Solution
> > An empty file with only the headers
> {: .solution} 
{: .challenge}
```

The other problem: context. Challenges in the `extra_challenges` file have to be tweaked to be self-sufficient because they will appear without any context. I think this challenge is a good example where such tweaks are obviously necessary.

maxim-belkin:
Yes, we ought to mention it somewhere :)

_extras/extra_challenges.md
```diff
@@ -8,12 +8,31 @@ permalink: /extra_challenges/
...
+> import pandas as pd
+> 
+> # Load the data into a DataFrame
+> surveys_df = pd.read_csv('data/surveys.csv')
+> 
+> # Loop through a sequence of years and print the year
+> start_year = 1977
+> end_year = 2002
+> for year in range(start_year, end_year+1):
+>     print(str(year))
```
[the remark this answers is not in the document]
wrightaprilm: Right, this is a function of the fact that where this challenge was embedded is right before we discuss creating new name strings to automate naming files. It can stay or go, in my opinion. Divorced from the context, perhaps it ought to go.

_extras/extra_challenges.md
```diff
@@ -0,0 +1,38 @@
...
+> ~~~
+> {: .language-python}
+>
+> What happens if there is no data for a year in a sequence? For example,
+> imagine we used `1976` as the start year in `range`
+>
+> > ## Solution
+> > An empty file with only the headers
```
wrightaprilm: @maxim-belkin and I just realized this part of the challenge doesn't make sense. It looks like some part of the challenge didn't get committed. I think this is meant to be what happens if you try to select rows that == 1976, in which case you will get an empty set of rows. Is that right?

NPalopoli: I think there could be two questions here. My reading of this challenge is, as there's no data for the year 1976 in the dataframe, what would be the output if you set `start_year = 1976`? Your suggestion of explicitly extracting rows where `year` equals `1976` would have a different output and thus could be a separate challenge.

_extras/extra_challenges.md
```diff
@@ -0,0 +1,38 @@
...
+> # Loop through a sequence of years and print the year
+> start_year = 1977
+> end_year = 2002
+> for year in range(start_year, end_year+1):
+>     print(year)
+> ~~~
+> {: .language-python}
+>
+> What happens if there is no data for a year in a sequence? For example,
+> imagine we used `1976` as the start year in `range`
```
wrightaprilm:
```suggestion
> imagine we used `1976` as the year in `surveys_df[surveys_df.year == year]`
```

I think you're exactly correct on what the challenge _meant_ to do. As it's written ... nothing special happens. You just start the `range` function one year earlier. I've made a suggestion here that makes the answer make sense. 

Sorry to make you do the labor on this as someone who just noticed the challenge was a little goofy;)

NPalopoli: Nice call @wrightaprilm, thanks a lot!

wrightaprilm:
I think all my review notes have been resolved. I think we can merge this and #344 if you're happy, @maxim-belkin.

maxim-belkin:
The code snippet, the question, and the answer are still not connected to each other.
It looks like the question is related to the following code snippet in the episode:
```python
# Load the data into a DataFrame                                                                     
surveys_df = pd.read_csv('data/surveys.csv')                                                         
                                                                                                     
for year in surveys_df['year'].unique():                                                             
                                                                                                     
    # Select data for the year                                                                       
    surveys_year = surveys_df[surveys_df.year == year]                                               
                                                                                                     
    # Write the new DataFrame to a CSV file                                                          
    filename = 'data/yearly_files/surveys' + str(year) + '.csv'                                      
    surveys_year.to_csv(filename) 
```

NPalopoli:
Thank you @maxim-belkin for making us notice. I have made some changes that should connect everything together. I hope it makes more sense now!

wrightaprilm:
Right, I think this was an ambiguous challenge to begin with. So now it's on us to decide what to do. The way I see it, we could 

1. Add code, as you've pasted in so there's context to simply add the whole challenge back.
2. Leave it as-is, and let the instructor decide.
3. ? Other options?

I don't feel strongly on which we want to do, and am happy to go along with whatever consensus emerges.

wrightaprilm:
Ah, I missed that comment. Huzzah, good work all!

Issue #360, opened 2019-02-07
Ep1: Better choice of dictionary keys


gcapes:
https://datacarpentry.org/python-ecology-lesson/01-short-introduction-to-Python/index.html#dictionaries

When introducing a dictionary, numeric keys are used. This confuses things slightly because it makes it seem that dictionaries can be indexed in the same way as lists, using numbers. I think it would be clearer to use something text-based for the first example.

maxim-belkin:
Thanks, Gerard! Very good issue that makes an important point.
Glad we fixed it before its first birthday :)

PR #370, opened 2019-04-05
Fixing confusion for indexing python dictionaries

Benfeitas:
See https://github.com/datacarpentry/python-ecology-lesson/issues/360

### DIFF
--- a/_episodes/01-short-introduction-to-Python.md
+++ b/_episodes/01-short-introduction-to-Python.md
@@ -369,12 +369,12 @@ a_list = [1, 2, 3]
 A **dictionary** is a container that holds pairs of objects - keys and values.
 
 ~~~
-translation = {'one': 1, 'two': 2}
+translation = {'one': 'first', 'two': 'second'}
 translation['one']
 ~~~
 {: .language-python}
 ~~~
-1
+'first'
 ~~~
 {: .output}
 
@@ -384,8 +384,8 @@ in the dictionary. Keys can only have particular types - they have to be
 "hashable". Strings and numeric types are acceptable, but lists aren't.
 
 ~~~
-rev = {1: 'one', 2: 'two'}
-rev[1]
+rev = {'first': 'one', 'second': 'two'}
+rev['first']
 ~~~
 {: .language-python}
 ~~~
@@ -394,7 +394,7 @@ rev[1]
 {: .output}
 
 ~~~
-bad = {[1, 2, 3]: 3}
+bad={['first','second','third']:'three'}
 ~~~
 {: .language-python}
 ~~~
@@ -410,13 +410,13 @@ user.
 To add an item to the dictionary we assign a value to a new key:
 
 ~~~
-rev = {1: 'one', 2: 'two'}
-rev[3] = 'three'
+rev = {'first': 'one', 'second': 'two'}
+rev['third'] = 'three'
 rev
 ~~~
 {: .language-python}
 ~~~
-{1: 'one', 2: 'two', 3: 'three'}
+{'first': 'one', 'second': 'two', 'third': 'three'}
 ~~~
 {: .output}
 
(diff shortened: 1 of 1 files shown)


pr4deepr:
Not sure if there is a confusion actually. 
The code:
`translation = {'one': 1, 'two': 2}`   and
`rev = {1: 'one', 2: 'two'}`
illustrate that both strings (one and two) and numeric types (1 and 2) are acceptable as keys..

RBenRocks:
@pr4deepr I agree with you. However, learners may think that indexing may be done in terms of key position and key order is not guaranteed to be preserved - aside from insertion order. For the lesson at hand, I think that https://github.com/datacarpentry/python-ecology-lesson/issues/360#issue-407755008 makes sense and is simpler if we just have it as text.

pr4deepr:
Thanks @RBenRocks . Agree with that, but I think this bit should not be modified, but you can add these changes later on instead? 
Towards the end of the 'dictionary' section, the issue you raise is addressed: 
It is important to note that dictionaries are “unordered” and do not remember the sequence of their items (i.e. the order in which key:value pairs were added to the dictionary). Because of this, the order in which items are returned from loops over dictionaries might appear random and can even change with time.
But, I think adding a line of code to address this would be good as it is an important concept.

maxim-belkin:
I've been thinking about this PR and issue #360 and couldn't understand their relationship because the issue  says:

> When introducing a dictionary, numeric keys are used.

and this PR changes values and, moreover, the very first dictionary introduced does not actually use numeric keys. Now I understand that we're talking about the `rev` dictionary... 

I support this change because we shouldn't be using numeric keys with dictionaries, actually:

```python
a = {1 : '2'}
print(a)
a[1.0] = '3'
print(a)
a[True] = '4'
print(a)
```

And we should definitely fix that `order in which key:value pairs were added to the dictionary` part because it is preserved in Python since recently. 

Also, statement `You can think about a key as a name for or a unique identifier for a set of values in the dictionary.` is not correct as there is one-to-one correspondence between keys and values, not one-to-many.

pr4deepr:
Thanks for clarifying this. 
Didn't realise the order was conserved recently..

_episodes/01-short-introduction-to-Python.md
```diff
@@ -444,16 +444,16 @@ for key in rev.keys():
...
+'second' -> two
+'third' -> three
 ~~~
 {: .output}
 
 > ## Changing dictionaries
 >
 > 1. First, print the value of the `rev` dictionary to the screen.
-> 2. Reassign the value that corresponds to the key `2` so that it no longer
+> 2. Reassign the value that corresponds to the key `second` so that it no longer
```
wrightaprilm: In this new set of assignments, I do agree that using strings is a little more explicit. But I think we should make it clear that dictionaries are not only for strings. Perhaps we should have them reassign the value of `second` to the numeral `2` instead of apple-sauce.

wrightaprilm:
I like this. I've made one comment on the challenge. If you agree, go ahead and add that to this PR, if not just let me know.

_episodes/01-short-introduction-to-Python.md
```diff
@@ -444,16 +444,16 @@ for key in rev.keys():
...
 >    reads "two" but instead "apple-sauce".
```
wrightaprilm:
```suggestion
>    reads "two" but instead "2".
```

wrightaprilm: I went ahead and added my comment as a suggestion, since you seem on board, @RBenRocks. If you click the "Commit suggestion" box, it'll add my change to your PR. Then I'll merge and we're done!

maxim-belkin:
A lot of things have changed since we discussed this PR. 
I will revive this PR with April's suggestions.

maxim-belkin:
Superseded by #445

PR #372, opened 2019-04-12
Use GitHub link instead of "DownGit" service

remram44:
I don't understand why this was done in #231. GitHub has built-in download links, I was surprised to see this fancy JavaScript service being used for this.

This seems like a 404 waiting to happen, and can't be good for accessibility.

[This is what DownGit itself uses anyway](https://github.com/MinhasKamal/DownGit/blob/def14cef2a5d9fe4bef3227c38de425e5c145c60/app/home/down-git.js#L152-L153):
```javascript
var downloadUrl = "https://github.com/"+repoInfo.author+"/"+
    repoInfo.repository+"/archive/"+repoInfo.branch+".zip";
```

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -9,7 +9,7 @@ title: Setup
 >
 > We will use the six files listed below for the data in this lesson.
 > Download these files to your computer either by clicking
-> [this link](https://minhaskamal.github.io/DownGit/#/home?url=https://github.com/weecology/portal-teachingdb),
+> [this link](https://github.com/weecology/portal-teachingdb/archive/master.zip),
 > which will give you everything in a single compressed file.
 > You'll need to unzip this file after downloading it.
 >


remram44:
The set of files appears to be different in either case, the ZIP has a bunch more files but is missing `speciesSubset.csv` and `bouldercreek_09_2013.txt`

wrightaprilm:
I'm hoping to get this fixed up for the new lesson release. It's a known issue that I've kind of punted on. All the lessons are meant to be using Figshare for data hosting, rather than storing in the repo. But maintainers don't have write access to that, so I'm waiting on someone in leadership to handle that.

I'm going to merge this for the time being, though, so that all the files are in sync. Thanks for noticing and sending us a PR, Remi.

PR #374, opened 2019-04-17
Removing unnecessary `

agbeltran:
In the instructor notes, there is an extra ` in one of the commands


maxim-belkin:
Thank you, Alejandra! 🎉

PR #375, opened 2019-04-17
Minor issue with word ordering

agbeltran:
"the both start bound and the stop bound" ---> "both the start bound and the stop bound"


maxim-belkin:
Thank you, Alejandra!

PR #380, opened 2019-05-17
01-short-introduction-to-Python.md: remove statement about order in Python dicts

maxim-belkin:
I suggest removing the statement about the order in Python dictionaries. The behavior has changed between Python 3.5, 3.6, and 3.7 and it would be a bit too difficult to explain.

### DIFF
--- a/_episodes/01-short-introduction-to-Python.md
+++ b/_episodes/01-short-introduction-to-Python.md
@@ -459,13 +459,6 @@ for key in rev.keys():
 >
 {: .challenge}
 
-It is important to note that dictionaries are "unordered" and do not remember
-the sequence of their items (i.e. the order in which key:value pairs were
-added to the dictionary). Because of this, the order in which items are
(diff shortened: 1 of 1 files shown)


wrightaprilm:
Oooh boy, how annoying. I'm not sure what to do here. If they follow our install instructions, they will download python 3.7, in which dicts are ordered. So, I guess my feeling is that we should just remove this, since this won't be the behavior going forward. I'm not sure if I think we should replace it with anything. Maybe something like:

"Like lists, dictionaries are ordered. However, ordered dictionaries are a recent innovation in python, and when you look for help on the internet or look at someone else's code from a published paper, you may find references to dictionaries being unordered."

maxim-belkin:
> Like lists, dictionaries are ordered

Yeah.... about that.... They preserve the insertion order but they aren't ordered. Let's, perhaps, just delete the paragraph.

Issue #389, opened 2019-05-26
possibly broken URLs


fmichonneau:
In preparation for the June 2019 release, we are looking for broken links in the lessons. We identified the links below as possibly being problematic:

```
 ● Resource: `setup`
    Message: ‘File referenced by URL doesn't exist.’
    Found in:
    - /home/francois/git/ecology-lessons/python-ecology/_site/index.html
 ● Resource: `http://ggplot2.org/`
    Message: ‘HTTP status code: 404’
    Found in:
    - /home/francois/git/ecology-lessons/python-ecology/_site/07-visualization-ggplot-python/index.html
  ● Resource: `https://spyder-ide.github.io`
    Message: ‘HTTP status code: 404’
    Found in:
    - /home/francois/git/ecology-lessons/python-ecology/_site/00-before-we-start/index.html
  ● Resource: `http://pandas.pydata.org/pandas-docs/stable/visualization.html#basic-plotting-plot`
    Message: ‘URL is valid but fragment (hash reference): 'basic-plotting-plot' not found in page.’
    Found in:
    - /home/francois/git/ecology-lessons/python-ecology/_site/02-starting-with-data/index.html
  ● Resource: `https://plotnine.readthedocs.io/en/stable/api.html#Color-and-fill-scales`
    Message: ‘URL is valid but fragment (hash reference): 'Color-and-fill-scales' not found in page.’
    Found in:
    - /home/francois/git/ecology-lessons/python-ecology/_site/07-visualization-ggplot-python/index.html
  ● Resource: `http://pandas.pydata.org/pandas-docs/stable/dsintro.html#dataframe`
    Message: ‘URL is valid but fragment (hash reference): 'dataframe' not found in page.’
    Found in:
    - /home/francois/git/ecology-lessons/python-ecology/_site/02-starting-with-data/index.html
  ● Resource: `https://docs.carpentries.org/topic_folders/policies/code-of-conduct.html#incident-reporting-guidelines`
    Message: ‘URL is valid but fragment (hash reference): 'incident-reporting-guidelines' not found in page.’
    Found in:
    - /home/francois/git/ecology-lessons/python-ecology/_site/CODE_OF_CONDUCT.html
```

Let me know if you have any questions. Thanks!

wrightaprilm:
[...]

PR #391, opened 2019-05-27
00-before-we-start.md: various fixes

maxim-belkin:
1. Fix link reported in #389 
2. Wrap text to 100 characters
3. Change formatting: `### text` to `**text:**` as these `###` are actually part of the same logical unit.


wrightaprilm:
Thanks, @maxim-belkin

PR #405, opened 2019-06-06
Small fixes (typos)

BinxiePeterson:


maxim-belkin:
Thank you, Bianca. All changes look good to me. My only suggestion would be
to keep ”Because” but I’ll let you and April make a decision on that.

wrightaprilm:
I could go either way. Since we're right before the release deadline, I'll merge as is to get it off our plates. Thanks, Bianca!

BinxiePeterson:
As I go through the rest of the lesson, I'll fix if needed, and I'll change "Since" back to "Because". When is the release?

maxim-belkin:
> When is the release?

Last Tuesday. But April used her magic wand to delay it.

PR #407, opened 2019-07-17
Update 03-index-slice-subset.md

miaomiao1011:
I wrote about the difference between two types of subsetting.


wrightaprilm:
Hi, sorry I'm so late in getting back with this. I like the idea of having the learners interact with the conceptual difference between the two methods of indexing. Is your idea that the learners will explain two each other, and identify situations in which they might want to use one or the other?

maxim-belkin:
Looks like there has been no activity for 5 months...

@miaomiao1011, I think the point you're trying to make is valid but we need to reformulate / reformat it a little bit. I suggest rephrasing your addition as a paragraph that conveys the same information in plain English without bullet points. What do you think?

Issue #361, opened 2019-02-07
Move solutions to exercises back to episodes


gcapes:
At least in the early episodes, most exercises don't have solutions.

maxim-belkin:
Solution are in the instructor notes... we need to move them

wrightaprilm:
As they are in the SWC-python? I'm fine with that. The thing to do about this might be to branch, and add all the solutions, then merge so there's a uniform roll-out.

maxim-belkin:
I meant here: https://datacarpentry.org/python-ecology-lesson/guide/

wrightaprilm:
Right, but you're proposing to move the answers out of there and into the lesson in a code fold?

maxim-belkin:
Oh, my apologies -- misread part of your previous response. Yes, this is exactly what I'm proposing (with appropriate modifications, of course).

wrightaprilm:
@maxim-belkin and I had a brief discussion on this, and we'd like to proceed like so:

1. We will adopt the swc-py style solutions format. See [here](https://github.com/swcarpentry/python-novice-inflammation/blob/gh-pages/_episodes/01-numpy.md) for an example, about 60% of the way down the page.

2. We would like one PR per lesson adopting this style. Each will be tested by us and merged once each lesson has one. 

3. We'd like this done by the summer lesson release (late June).

This is a beginner-friendly activity and would be great for someone who is interested in learning a little more about how the lesson template works. If no one wants to try, Max and I will handle it prior to the release.

btovar:
Fixed in #542

PR #408, opened 2019-07-29
added examples to the first two episodes of python ecology class from…

ezgirmak:
Hi,

As part of my instructor checkout procedure, I decided to embed examples back into the lesson materials from the instructor notes as discussed in the issue #361. I did the first two episodes and tried to follow carpentries style as much as possible. If everything looks good, I can continue to do the rest of the episodes. 

Here are some issues related to the second episode. 

- Second challenge in grouping challenges has a further example in the instructor notes 'I want to know the last survey year, median foot-length and mean weight for each plot/sex combination' Should I add it as an additional challenge? 

-  The instructor notes and the example in the lesson for Calculating Statistics Challenges is not the same. In the instructor notes, the name of the variable is `plot_names`, in the example it is ` site_names`. The code to get  number of unique species in the data in the instructor notes is `len(pd.unique(surveys_df["species"]))` and there is no column named `species`.  So I wrote it as 
`len(pd.unique(surveys_df["species_id"]))` in the lesson solutions. I have not changed the notes but I can change them in my next commit.

Thanks,
Ezgi

### DIFF
--- a/_episodes/01-short-introduction-to-Python.md
+++ b/_episodes/01-short-introduction-to-Python.md
@@ -361,6 +361,16 @@ a_list = [1, 2, 3]
 > 2. What happens when you execute `a_tuple[2] = 5`?
 > 3. What does `type(a_tuple)` tell you about `a_tuple`?
 >
+> > ## Solution
+> >
+> >As a tuple is immutable, it does not support item assignment. Elements in a list can be altered individually.
+> >
+> >
+> > ~~~
+> > tuple
+> > ~~~
+> > {: .output}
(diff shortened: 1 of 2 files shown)


maxim-belkin:
Thank you, Ezgi! To speed up the review/discussion process, could you please [split your pull request](https://github.com/swcarpentry/python-novice-inflammation/wiki/Splitting-(Breaking-down)-Pull-Requests) into smaller ones (solution-per-PR)? I linked a quick guide on how to do that but please let me know if you experience any difficulties along the way.

PR #412, opened 2019-08-01
Re-use previously introduced variables

katrinleinweber:
Hello :-) I wondered whether it was intentional that only 2 of the 3 initially introduced variables are reused in the immediately following examples. If not, here's a suggestion to do just that.


wrightaprilm:
Good catch! Thanks!

katrinleinweber:
You're welcome, and thanks for merging :-)

Issue #358, opened 2019-02-04
Ep0: SciPy keypoint not covered in lesson text


gcapes:

wrightaprilm:
Good catch. If anyone is looking for an instructor training checkout exercise, revising the keypoints and learning objectives on 00 would be a good one.

PR #415, opened 2019-08-07
Jensentobias/fixing 00 objectives and keypoints

jensentobias:
Motivation
- removed SciPy keypoint as mentioned in #358 
- Revised objectives (also mentioned in #358) based on current content and reordered to match current content


wrightaprilm:
This looks good to me if it looks good to you, @maxim-belkin.

maxim-belkin:
Looks great! Thanks, @jensentobias.

PR #417, opened 2019-08-21
added instructions on shebang

### LINKED ISSUE #311: Suggestions for Short Introduction to Programming in Python
For example.py is it worth showing a shebang at the top of the file and explaining what it means?

In the Python in the Short Intro Indentation section it would be helpful to include that indentation is achieved by either hitting the tab key or three spaces but not both within one script.

echism:
fixes #311

### DIFF
--- a/_episodes/01-short-introduction-to-Python.md
+++ b/_episodes/01-short-introduction-to-Python.md
@@ -186,6 +186,39 @@ Notice that "Data Carpentry" is printed only once.
 lesson, we will introduce methods and user-defined functions. The Python
 documentation is excellent for reference on the differences between them.
 
+*Making the script executable*
+
+To specify that example.py is a python script, we can add something called the shebang to our script. It looks like this:
+
+*example.py*
+~~~
+#!/usr/bin/env python3
+# The previous line is the shebang!
+# A Python script file
+# Comments in Python start with #
+# The next line assigns the string "Data Carpentry" to the variable "text".
+text = "Data Carpentry"
+
+# The next line does nothing!
+text
+
+# The next line uses the print function to print out the value we assigned to "text"
+print(text)
+~~~
+{: .language-python}
+
+Once we've added the shebang, we can now run script like this:
+
+~~~
+$ ./example.py
+~~~
+{: .language-bash}
+
+~~~
+Data Carpentry
+~~~
+{: .output}
+
 ### Operators
 
 We can perform mathematical calculations in Python using the basic operators


_episodes/01-short-introduction-to-Python.md
```diff
@@ -186,6 +186,39 @@ Notice that "Data Carpentry" is printed only once.
...
+
+# The next line does nothing!
+text
+
+# The next line uses the print function to print out the value we assigned to "text"
+print(text)
+~~~
+{: .language-python}
+
+Once we've added the shebang, we can now run script like this:
```
maxim-belkin: You also need to `chmod +x example.py`...

maxim-belkin:
Thanks for the PR, @echism! Could you please justify the need for adding this note to the "Short introduction to Python"?

echism:
It’s not strictly necessary, but it is easier to invoke a script this way. It’s more of a tip, one suggested in the issues.

Elizabeth Prout
Analyst/Programmer
UITS Advanced Visualization Lab
https://showcase.avl.iu.edu/projects

Indiana University
University Information Technology Services
535 West Michigan Street
Indianapolis, IN 46202
317-278-3647

The Advanced Visualization Lab at Indiana University is a unit of the Research Technologies division of UITS; Research Technologies is a PTI Cyberinfrastructure and Service Center.

wrightaprilm:
If they followed the installation instructions, they should have Anaconda Python in their path. If they followed the install instructions, for example:

```python
import sys
print(sys.version)
```

will return:

```
3.6.8 |Anaconda custom (64-bit)| (default, Dec 29 2018, 19:04:46) 
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]
```

Or similar without the shebang. You'll note this in `_includes/scripts`, for example. This PR straddles the line of what ought to be in the lesson, and not. I don't really want to put us in a position of maintaining instruction sets for multiple Python versions. If we want to include this, I think the best course of action would be to make a callout box noting that self-study students or instructors using custom Python installs should check the instructor guide. Then, this information could go in `_extras/guide.md`.

maxim-belkin:
@wrightaprilm, @echism is talking about something else: one can "execute" Python scripts (as in`./my-script.sh`) if they:
1. add `#!/usr/bin/env python3` as the first line in the script and
2. give this file executable permissions.

There shouldn't be be any burden on the maintainers. I was more concerned with the fit of this material to the lesson in its current state. While I agree that it _might_ be convenient -- personally, I don't use it that much (I prefer to `python3 script.py` to be on a safe side in terms of which python interpreter I use). Therefore, I'm not even sure about the callout box. Perhaps, this should be as an extra information for instructors.

maxim-belkin:
Unfortunately, I'm going to pass on this, @echism. Shebangs are convenient but they would be a better fit for a final episode on Python (a cherry on top sort of thing) rather than the first one. Thank you for your contribution anyway! I'm glad your PR stimulated this discussion -- it will help myself and future maintainers of the lesson.

PR #420, opened 2019-09-15
Fix typo that shows a wrong value of the `number` variable.

ielis:
There is a typo/error within the file. The value of the `number`
variable is set to `42` on line 84. However, the line `151` shows `11`
instead of `42`. This might be confusing to people/instructors.


maxim-belkin:
Thanks, Daniel!

maxim-belkin:
The bug was introduced in #412 when we switched from using (hard-coded) numbers to variables.

PR #422, opened 2019-10-01
Show popular SO questions, instead of newest

katrinleinweber:
Same as https://github.com/swcarpentry/python-novice-gapminder/pull/443 ;-)


maxim-belkin:
Thank you, Katrin!

katrinleinweber:
You're welcome, and thanks for merging :-)

PR #430, opened 2019-12-10
fix formatting

maneesha:
`md` formatting made the sentence ending "located at position 0." read as beginning an ordered list.  This PR fixes that formatting.

### DIFF
--- a/_episodes/03-index-slice-subset.md
+++ b/_episodes/03-index-slice-subset.md
@@ -113,8 +113,9 @@ the related Python data type dictionary).
 {: .callout}
 
 Let's remind ourselves that Python uses 0-based
-indexing. This means that the first element in an object is located at position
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Thanks, Maneesha. I think we should keep these sentences in a single paragraph. Could you please make each sentence appear on its own line instead?

> Let's remind ourselves that Python uses 0-based indexing.
This means that the first element in an object is located at position 0.
This is different from other tools like R and Matlab that index elements
within objects starting at 1.

maxim-belkin:
Maneesha, may I also ask you to use your personal forks of the lessons for submitting pull requests? I think it would be best if we continue to set good example for our contributors and follow the same procedures for proposing changes to the lessons as them despite having access privileges.

maxim-belkin:
fixed in #444.

PR #431, opened 2019-12-25
Link to new conda homepage

katrinleinweber:
See diff ;-)

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -44,7 +44,7 @@ title: Setup
 ## Install the workshop packages
 
 For installing these packages we will use Anaconda or Miniconda.
-They both use [Conda](http://conda.pydata.org/docs/), the main difference is
+They both use [Conda](https://conda.io/en/latest/), the main difference is
 that Anaconda comes with a lot of packages pre-installed.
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Thanks, Katrin. When I go to https://conda.io it redirects me to https://conda.io/en/latest
Should we use this shorter url?

katrinleinweber:
Exactly to shave off the redirect time, was why I'd suggest the exact URL ;-)
If they ever change it, they seem likely to redirect it as well.

maxim-belkin:
Hmmm, I was not concerned with redirection time but with pointing to "/en/" site, _but_... I have to agree with you here for several reasons: 

1. in my tests (I used http://www.serp-perception.com) redirects were taking 50% of page loading time
2. redirection was taking close to a second (~ 0.7 - 0.8 seconds), which is... not nice
3. The website uses 302 redirect, which is a temporary redirect. It means they might change it in the future or even remove it completely.

Thanks, Katrin!

katrinleinweber:
You're welcome & thanks measuring! Good to know that female intuition was confirmed here ;-D

PR #432, opened 2020-01-06
Update 01-short-introduction-to-Python.md

yrad:
Short elaboration on what "hashable" means. If "hashable" is used to explain a concept it needs to be better communicated. Without elaboration it is an example of minimal guidance and could better be removed.  A short elaboration makes it possible to pose the question "Why not lists".

### DIFF
--- a/_episodes/01-short-introduction-to-Python.md
+++ b/_episodes/01-short-introduction-to-Python.md
@@ -379,7 +379,7 @@ translation['one']
 
 Dictionaries work a lot like lists - except that you index them with *keys*.
 You can think about a key as a name for or a unique identifier for a set of values
-in the dictionary. Keys can only have particular types - they have to be
+in the dictionary. Keys can only have particular types that are not mutable or likely to change - they have to be
 "hashable". Strings and numeric types are acceptable, but lists aren't.
 
 ~~~


maxim-belkin:
Thank you, Duncan. I agree with you in that we should either explain what "hashable" is or not mention it at all. Provided that this is a short intro to Python, I suggest we do the latter. Are you up for that? Alternatively, we could rewrite the whole paragraph/section, because it is best not to use integers/floating-point numbers as keys...

yrad:
For sure. The gist of the lesson about dictionaries is that a dictionary is unordered, judging by the 'key points' summary and instructor notes. Happy to have the hashable bit removed. Is there room for a sentence about why we'd use a dictionary over a list? Something more than introducing it as a vehicle for a traceback error?

maxim-belkin:
> For sure. The gist of the lesson about dictionaries is that a dictionary is unordered, judging by the 'key points' summary and instructor notes.

Let's update the lesson and instructor notes to remove the "unordered" piece. While dictionaries are unordered (that is, if `a` is a dict with keys `"one"` and `"two"` and `b` is a list with two elements, `a["one"]` is not connected to `a["two"]` as b[0] is to b[1]), insertion order in dictionaries is preserved (recent change) and us saying that dictionaries are unordered has already confused quite a few people.

> That said, it does reappear in the traceback error. Is the dictionary introduced as a vehicle for the traceback error?

No. We can and, perhaps, should remove the traceback error itself. What do you think?

yrad:
> Let's update the lesson and instructor notes to remove the "unordered" piece. 

Will do.
> insertion order in dictionaries is preserved (recent change) 

I've learned something :)
> remove the traceback error

Sure, the Tuples vs. Lists exercise generates a traceback error so removing the cictonary traceback would not erase errors entirely. I'll modify the pull request.

maxim-belkin:
> Sure, the Tuples vs. Lists exercise generates a traceback error so removing the dictonary traceback would not erase errors entirely. I'll modify the pull request.

Thank you. Let's update the episode incrementally. It is always better to have several small PRs than a few big ones.

PR #433, opened 2020-01-12
Unordered dictionaries

yrad:
Remove unordered from dictionary definition in keypoint summary to limit confusion

### DIFF
--- a/_episodes/01-short-introduction-to-Python.md
+++ b/_episodes/01-short-introduction-to-Python.md
@@ -378,8 +378,7 @@ translation['one']
 {: .output}
 
 Dictionaries work a lot like lists - except that you index them with *keys*.
-You can think about a key as a name for or a unique identifier for a set of values
-in the dictionary. 
+You can think about a key as a name or unique identifier for the value it corresponds to. 
 ~~~
 rev = {1: 'one', 2: 'two'}
 rev[1]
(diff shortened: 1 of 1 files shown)


_episodes/01-short-introduction-to-Python.md
```diff
@@ -379,9 +379,7 @@ translation['one']
 
 Dictionaries work a lot like lists - except that you index them with *keys*.
 You can think about a key as a name for or a unique identifier for a set of values
-in the dictionary. Keys can only have particular types - they have to be
-"hashable". Strings and numeric types are acceptable, but lists aren't.
-
+in the dictionary.
```
maxim-belkin: 1. In a Python dictionary, a key corresponds to a single value only (that can be a list or other complex data structure but nonetheless...)
2. Repetitive "for": 
> as a name ~for~ or a unique identifier for
3. I suggest removing the last `in the dictionary`.

yrad: Thanks Maxim, I'll add these suggestions but I worry that a change to:

> You can think about a key as a name or unique identifier for a set of values. In a Python dictionary, a key corresponds to a single value only.

might introduce some confusion about pairing i.e. between a key and a "set of values" or a "single value" . Would:

> You can think about a key as a name or unique identifier for the value it corresponds to. In a Python dictionary, a key corresponds to a single value only.

be an improvement?

yrad: Thanks Maxim, I'll add these suggestions but I worry that a change to:

    You can think about a key as a name or unique identifier for a set of values. In a Python dictionary, a key corresponds to a single value only.

might introduce some confusion about pairing i.e. between a key and a "set of values" or a "single value" . Would:

    You can think about a key as a name or unique identifier for the value it corresponds to. In a Python dictionary, a key corresponds to a single value only.

be an improvement?

yrad: In hindsight, maybe that's what you'd meant in suggestion one.

maxim-belkin: > In hindsight, maybe the improvement I proposed is what you'd meant in suggestion one.

Yes, exactly. The statement we previously had in the document was incorrect because there is a clear one-to-one correspondence between keys and values.

PR #438, opened 2020-01-18
Add a reminder about what the question mark does

maneesha:


maneesha:
Catching up on a very old issue!
I am fine with any of the options listed. I originally put in this PR because the `?` notation is used in the example, and so the explanation should refer to what is in the example.

tobyhodges:
I'm helping the current lesson Maintainers process outstanding pull requests on this repository, in preparation for [transition to the new lesson infrastructure](https://carpentries.github.io/workbench/faq.html#update-fork-from-styles). Thank you for contributing, @maneesha. 

I agree that the lesson could be improved with an explanation of `?`, given that it is used in the example code block. Of the three options @maxim-belkin listed, I would be most in favour of the second ("Expand the sentence to mention the built-in help() function + clarify that in Jupyter notebooks we can use topic? notation"), so I am going to suggest a change to the PR.

We have until the end of this month to get these changes merged, before the Workbench transition forces us to close the PR.

PR #435, opened 2020-01-17
Assorted minor fixes and improvements

katrinleinweber:
As I'm preparing [a workshop](https://tibhannover.github.io/2020-02-13-LSE-Hannover/) with this lesson, I'll collect some typo fixes, formatting improvements, and IMHO better wordings here. Please feel free to review or cherry-pick already.

PR #453, opened 2020-02-24
guide.md: remove space after comma in 'Basic Loop Challenges'

maxim-belkin:
Have to undo one recent change introduced in #435 in order to keep the section consistent with what it was doing before. Also, while at it, add spaces around `+`.

@katrinleinweber, LGTY?


katrinleinweber:
I'm not maintainer here, but ` + ` is definitely good style :-)
`','` or  `', '` is IMHO purely a matter of preferences in the output formatting.
Either is fine with me.

maxim-belkin:
Thanks, Katrin.

Issue #454, opened 2020-02-25
03-index-slice-subset.md: wrong operator shown in Objectives (`=` instead of `==`)


snownontrace:
The 7th item of the Objectives of this lesson (https://datacarpentry.org/python-ecology-lesson/03-index-slice-subset/) states:

"Query / select a subset of data using a set of criteria using the following operators: =, !=, >, <, >=, <=."

The first operator should be "==" rather than "=."

maxim-belkin:
Yes, you're correct, Shaohe. Would you be willing to submit a PR to fix that?

snownontrace:
Yes! Plus a couple of additional typos I spotted. Will do some time in this
week.

Best,
Shaohe

maxim-belkin:
Awesome, thank you!

PR #455, opened 2020-02-26
Corrected operator and estimated timing

snownontrace:
For Issue #454
The wrong operator shown "=" has been corrected to "==".


maxim-belkin:
Makes sense! Thank you Shaohe!

PR #459, opened 2020-05-06
Update 02-starting-with-data.md

abdulanl:
There are an error in the Statistics Challenges, as it defers from the instructor notes. Also, the idea behind the exercise is to get same result from both methods, while indicates the easiness of using the `unique` method

### DIFF
--- a/_episodes/02-starting-with-data.md
+++ b/_episodes/02-starting-with-data.md
@@ -395,11 +395,11 @@ array(['NL', 'DM', 'PF', 'PE', 'DS', 'PP', 'SH', 'OT', 'DO', 'OX', 'SS',
 
 > ## Challenge - Statistics
 >
-> 1. Create a list of unique site ID's ("plot_id") found in the surveys data. Call it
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Hi Abdula!

Thank you for contributing to the lesson. [Our data](https://datacarpentry.org/python-ecology-lesson/02-starting-with-data#our-data) section describes`plot_id` column as "ID of a particular site". So, I think it makes sense to keep using the word "site" when talking about elements of that column. The next question is whether variable `site_names` is reasonable or if we should switch to `plot_names`. I think both variable names are reasonable. If you feel that `site_names` is not a proper name, I suggest rephrasing the second sentence to:

> You may call it `plot_names` or `site_names`.

What do you think?

tobyhodges:

Thank you for contributing, @abdulanl. I am going to close this pull request due to inactivity. However, if you wish to return and address the comments made by the former Maintainer above, please feel free to reopen and update it before the end of this month, or open a new PR after the infrastructure transition has taken place in May.

PR #464, opened 2020-06-16
Add code to look at the head from true_copy_surveys_df

ricdea:
Added a few lines to look at the head from the copied data frame.

This ensures students see that the copy has not changed.

Otherwise the object is created but not used.

### DIFF
--- a/_episodes/03-index-slice-subset.md
+++ b/_episodes/03-index-slice-subset.md
@@ -240,12 +240,15 @@ Let's try the following code:
 # ref_surveys_df was created using the '=' operator
 ref_surveys_df.head()
 
+# true_copy_surveys_df was created using the copy() function
+true_copy_surveys_df.head()
+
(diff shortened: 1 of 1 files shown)


_episodes/03-index-slice-subset.md
```diff
@@ -254,6 +257,8 @@ Remember we created the reference `ref_survey_df` object above when we did
 refer to the same exact DataFrame object. If either one changes the object,
 the other will see the same changes to the reference object.
 
+However - `true_copy_surveys_df` was created via the `copy()` function. It retains the original values for the first three rows.
```
maxim-belkin, line 260: Could you please break this line so that it's less than 100 characters long? I suggest breaking after the period.

tobyhodges:

Thank you for contributing, @ricdea. I have gone ahead and broken up the long line as requested by Maxim above, and will now merge this PR.

PR #491, opened 2021-04-13
Fix questions introduction episode

svenvanderburg:
* Fix questions 1st episode (introduction to python programming) (the old questions were a copy from the 0th episode that is an introduction to python and why we should learn it, I changed them to reflect the content of the episode)
* Add time for challenges

### DIFF
--- a/_episodes/01-short-introduction-to-Python.md
+++ b/_episodes/01-short-introduction-to-Python.md
@@ -3,8 +3,8 @@ title: Short Introduction to Programming in Python
 teaching: 30
 exercises: 5
 questions:
-    - "How do I program in Python?
-    - "How can I represent my data in Python?
+    - "How do I program in Python?"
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Thanks for the PR, @svenvanderburg. The changes you suggested make perfect sense to me. I do wonder, however, if we should keep the first question ("What is Python?"). I can't say that I'm a big fan of how this question is answered in this episode, but it strikes me as the right question to ask in the beginning of the lesson. What do you think?

svenvanderburg:
@maxim-belkin I realized that my description of what changed is a tad short. I have updated it now. I hope it is clear that the questions from episode 1 were a copy from episode 0. The question 'What is Python?' is already answered in episode 0 and in this (episode 1) episode we introduce programming in python.

maxim-belkin:
Makes sense! Thank you, @svenvanderburg!

Issue #490, opened 2021-04-09
Problems with check_env.py


MathewBiddle:
In preparing for my teaching demo next week :flushed: I've been reviewing the [instructor notes](https://datacarpentry.org/python-ecology-lesson/guide/index.html) for this lesson and came across two problems with [check_env.py](https://github.com/datacarpentry/python-ecology-lesson/blob/gh-pages/_includes/scripts/check_env.py).

1. When I run `python check_env.py` it fails with the message  `AssertionError: Anaconda is NOT installed as your default version of Python. Please make sure that is in accordance with the instructions provided.`
   1. I think this can be resolved through a PR by adjusting https://github.com/datacarpentry/python-ecology-lesson/blob/7a2319fd3d15f04608e999010904b2a7a485a50c/_includes/scripts/check_env.py#L18 to indicate `Anaconda` (note the capital A). As that is now how the structure is created with the Anaconda distribution. There is probably a cleaner way to do this, however I think that will fix for now. 
1. Why is `ggplot` required for this workshop? Shouldn't that be `plotnine`? https://github.com/datacarpentry/python-ecology-lesson/blob/7a2319fd3d15f04608e999010904b2a7a485a50c/_includes/scripts/check_env.py#L27

I can submit a PR to make these changes, if they are appropriate.

As a side note, it might be helpful to provide an [environment.yml](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#creating-an-environment-from-an-environment-yml-file) file in the [setup](https://datacarpentry.org/python-ecology-lesson/setup.html) as well. I looked around and couldn't find anything (although there is a lot of material to cover here, so I might have missed it). If it's of interest, this is what I have so far:
```
name: python-ecology-lesson
channels:
  - conda-forge
  - defaults
dependencies:
  - python=3.8
  - jupyter
  - pandas
  - numpy
  - matplotlib
  - plotnine
```
Which could be built with `$ conda env create -f .\environment.yml`

maxim-belkin:
Thanks for the report, Mathew. I think we're dealing with a few issues here:

1. The "Instructor Notes" page isn't the right place for this kind of information. Perhaps, we should move it to `setup.md`.
2. `_includes/scripts` isn't the right place for this script either (it's hard to find). If we want participants to use this script, it should be provided in the ZIP file that we provide on the setup page.
[...]

PR #493, opened 2021-04-30
adding environment.yml and instructions

MathewBiddle:
Addressing parts of #490

* I've added an environment.yml file to the root directory (wasn't sure where to put it exactly).
* I updated the setup.md to include information on how to use conda to create the environment file referenced above.
* Adjusted the check_env.py script to check for plotnine and not ggplot. Also, fixed the check with newer anaconda installs using capital A, Anaconda path.

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -169,7 +169,7 @@ conda install -c conda-forge plotnine
 ~~~
 {: .language-bash}
 
-### _(Optional)_ Installing with environment file
+### _(Alternative)_ Installing required packages with environment file
 Download the 
 [environment.yml](https://raw.githubusercontent.com/datacarpentry/python-ecology-lesson/gh-pages/environment.yml) 
 file by right-clicking the link and selecting save as.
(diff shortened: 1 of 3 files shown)


setup.md
```diff
@@ -169,6 +169,25 @@ conda install -c conda-forge plotnine
 ~~~
 {: .language-bash}
 
+### _(Optional)_ Installing with environment file
```
maxim-belkin, line 172: It might be better to say "Alternative" rather than "Optional" and mention "Required packages" somewhere to make it clear that we're talking about installing required packages, not Miniconda/Anaconda

setup.md
```diff
@@ -169,6 +169,25 @@ conda install -c conda-forge plotnine
...
+{: .language-bash}
+
+Activate the new environment with:
+~~~
+conda activate python-ecology-lesson
+~~~
+{: .language-bash}
+
+For more information on environment files, see the conda documentation for 
+[creating an environment from an environment.yml file](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#creating-an-environment-from-an-environment-yml-file).
```
maxim-belkin, line 190: This is a good reference but I wonder if learners will benefit from it.

Shall this method also mention `conda deactivate`?

maxim-belkin (commented):
Thanks, @MathewBiddle! It looks really good. I do have a few questions/concerns that I'd like to get clarified/fixed before merging.

maxim-belkin:
Thanks a lot for your contribution, @MathewBiddle! I think this is a very fine addition to the setup instructions.

Issue #496, opened 2021-09-13
Ep3 - Mislabelled cells


nadiamuhe:
There are two cells are that labelled "Python" instead of "Output". This can be confusing to new learners. 

To be specific, the cells contain the following output: 

1)
```
array(['NL', 'DM', 'PF', 'PE', 'DS', 'PP', 'SH', 'OT', 'DO', 'OX', 'SS',
       'OL', 'RM', nan, 'SA', 'PM', 'AH', 'DX', 'AB', 'CB', 'CM', 'CQ',
       'RF', 'PC', 'PG', 'PH', 'PU', 'CV', 'UR', 'UP', 'ZL', 'UL', 'CS',
       'SC', 'BA', 'SF', 'RO', 'AS', 'SO', 'PI', 'ST', 'CU', 'SU', 'RX',
       'PB', 'PL', 'PX', 'CT', 'US'], dtype=object)
```
2)
```
        record_id     month        day         year    plot_id  \
sex
F    18036.412046  6.583047  16.007138  1990.644997  11.440854
M    17754.835601  6.392668  16.184286  1990.480401  11.098282

     hindfoot_length     weight
sex
F          28.836780  42.170555
M          29.709578  42.995379
```

maxim-belkin:
@nadiamuhe, would you like to submit a PR with a fix?

nadiamuhe:
@maxim-belkin Sure! I will submit one.

PR #499, opened 2021-09-13
#496 Fixed mislabelled cells.md

nadiamuhe:
Issue #496 : Fixed mislabelled cells.


maxim-belkin (approved):
👍🏻 Thanks, @nadiamuhe!

PR #500, opened 2021-09-17
Delete .travis.yml

fmichonneau:
I would suggest to remove .travis.yml given that we have now robust GitHub Actions to take care of the Continuous integration.


maxim-belkin (approved):
Good bye, Travis. You have served us well.

PR #507, opened 2022-05-12
Using display() command instead of print() in 02-starting-with-data and added assignment of data types

wgiese:
(a) I would suggest to introduce the use of the display() function from the IPython package, which is preferred over print() for pandas. display() give essentially the same output as typing a pandas data frame name into the Python command prompt or at the end of a code block, but it is useful for longer code blocks. 
(b) I would  suggest to explain that surveys_df.dtypes can be easily adjusted by one line of code, for instance for the 'sex' column, a categorical datatype might be preferred instead of 'object', or 'int16' instead of 'int64' for the columns 'month' or 'year'.
I hope this can be useful, let me know what you think.

### DIFF
--- a/_episodes/02-starting-with-data.md
+++ b/_episodes/02-starting-with-data.md
@@ -534,7 +534,7 @@ ways, but we'll use `groupby` combined with **a `count()` method**.
 ~~~
 # Count the number of samples by species
 species_counts = surveys_df.groupby('species_id')['record_id'].count()
-print(species_counts)
+display(species_counts)
 ~~~
 {: .language-python}
 
--- a/_episodes/02-starting-with-data.md
+++ b/_episodes/02-starting-with-data.md
@@ -317,6 +317,30 @@ fractional values, but the weight and hindfoot_length columns can, because they
 have type `float64`. The `object` type doesn't have a very helpful name, but in
 this case it represents strings (such as 'M' and 'F' in the case of sex).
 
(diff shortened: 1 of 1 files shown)


tobyhodges:

Thank you for suggesting these changes, @wgiese. For future reference when making multiple, unrelated suggestions such as these, it is better to open separate pull requests for each suggestion. That being said, I will respond to each one here:

a) While I agree that `IPython.display.display` is useful, I would prefer not to introduce another import into the lesson.

b) I like this suggestion very much, especially in relation to the conversion of `sex` to the more specific and appropriate `categorical` type. However, there already exists [a later episode on data types and formats](https://datacarpentry.org/python-ecology-lesson/04-data-types-and-format/index.html), which would be the more appropriate place for this content. I note that that episode does not mention the `categorical` datatype, and I would welcome a new pull request adding this, along with your example of how to simultaneously adjust the type of all the columns in the dataframe.

I am going to close this now, but I do encourage you to follow up with a new PR.

An important final comment: if you have the time and inclination to return to this lesson almost one year later and open a new pull request, please be aware that we will transition this lesson (and all others!) to the new infrastructure at the beginning of May 2023. Your new PR will need to be merged by then, or it will be invalidated and closed during the transition. So you may prefer to wait until the transition is done, then contribute to the updated version of the lesson. Either way, I would be delighted to help you make the change.

Issue #528, opened 2023-03-01
locate `_include/scripts`


goi42:
The [Instructor Notes](https://datacarpentry.org/python-ecology-lesson/guide/) for [Data Analysis and Visualization in Python for Ecologists](https://datacarpentry.org/python-ecology-lesson/index.html) refers to something called "the `_include/scripts` directory". What is this? I do not find it mentioned on the [setup page](https://datacarpentry.org/python-ecology-lesson/setup.html) or in the [zipped files](https://datacarpentry.org/python-ecology-lesson/data/portal-teachingdb-master.zip).

tobyhodges:
Thanks for raising this, @goi42. The directory is located [in this lesson repository](https://github.com/datacarpentry/python-ecology-lesson/tree/gh-pages/_includes/scripts). If you would like to open a PR to link to its location on GitHub from the Instructor Notes page, I would be delighted to review it.

btovar:
Fixed in #530

PR #530, opened 2023-03-02
link to _includes/scripts/

goi42:
First reported as #528.

### DIFF
--- a/_extras/guide.md
+++ b/_extras/guide.md
@@ -11,7 +11,7 @@ encounter setup issues, please file an issue with the tags 'High-priority'.
 
 ## Checking installations.
 
-In the [`_includes/scripts`](../_includes/scripts) directory, you will find a script called check_env.py This checks the
+In the [`_includes/scripts`](https://github.com/datacarpentry/python-ecology-lesson/tree/gh-pages/_includes/scripts) directory, you will find a script called check_env.py This checks the
 functionality of the Anaconda install.
 
 By default, Data Carpentry does not have people pull the whole repository with all the scripts and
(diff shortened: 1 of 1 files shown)


goi42:
@tobyhodges request for review

_extras/guide.md
```diff
@@ -11,12 +11,12 @@ encounter setup issues, please file an issue with the tags 'High-priority'.
 
 ## Checking installations.
 
-In the `_include/scripts` directory, you will find a script called check_env.py This checks the
+In the [`_includes/scripts`](../_includes/scripts) directory, you will find a script called check_env.py This checks the
```
tobyhodges, line 14: I'm not sure this link will work in the lesson site after the page has been built and rendered - I think it will end up pointing to https://datacarpentry.org/python-ecology-lesson/_includes/scripts, which does not exist.

Anyway, I think the link target is best set as the full URL to the directory on GitHub, i.e. "https://github.com/datacarpentry/python-ecology-lesson/tree/gh-pages/_includes/scripts" because this is most reliable way for someone to find it when teaching a workshop. 

If they are working offline, they will need to have downloaded the repository and built their own local version. In which case I suspect they are likely to also be able to find the local equivalent location of the script.

tobyhodges (changes requested):
Thank you, @goi42. I would like to use a different target URL for the link but otherwise this is good to go. Great that you also caught a couple of typos.

tobyhodges (approved):

tobyhodges:
Thank you @goi42 🙌

PR #535, opened 2023-03-05
correct description of add_axes parameters

goi42:


_episodes/08-putting-it-all-together.md
```diff
@@ -204,7 +204,7 @@ ax1.set_xlabel('value')
 
 # add additional axes to the figure
 ax2 = fig.add_axes([0.125, 0.575, 0.3, 0.3])
-#ax2 = fig.add_axes([left, bottom, right, top])
+#ax2 = fig.add_axes([left, bottom, width, height])
```
vinisalazar, line 207: I'm wondering if we should simply remove this line since it's commented?

goi42: I think it’s useful to show how the command works (call signature of the function).

vinisalazar: That's understandable, but I don't think it's good practice/readable to have a commented bit of code like this.

Moreover, IMHO `left` and `bottom` don't really match with `width` and `height`, whereas `right` and `top` do (even if that's what's actually in the function docstring).

I would suggest an in-line comment like (note the double space preceding the comment):

```python
ax2 = fig.add_axes([0.125, 0.575, 0.3, 0.3])  # number coordinates correspond to left, bottom, width, heigh
```

OR

```python
left, bottom, width, height = 0.125, 0.575, 0.3, 0.3
ax2 = fig.add_axes([left, bottom, width, height])
```

Although this second example uses variable unpacking and that might be a new concept that isn't previously introduced in the lesson.

vinisalazar: An alternative could also be:

```python
rectangle_coordinates = [0.125, 0.575, 0.3, 0.3]
fig.add_axes(rectangle_coordinates)
```

I think this one is really readable and also matches the function's docstring instructions.

goi42: It’s true this is a bit ugly, and this (I think) is the only place with in-line documentation in this format. It would be no great loss to remove it.

tobyhodges: agreed. @goi42 would you be willing to update the PR to use @vinisalazar's suggested change adding the in-line comment?

LilithElina:
Thank you, @goi42! If nobody minds, I'd like to add the "plot beta distribution" comment to the updated version as well.

tobyhodges:
Sure, @LilithElina. I note that this PR now has conflicts that need to be resolved - @goi42 or @LilithElina if you can address those, I think this one is will be ready to merge too.

PR #536, opened 2023-03-05
better document overlaid beta distribution

goi42:


vinisalazar (approved):
LGTM, thank you @goi42 !

PR #540, opened 2023-04-14
Update to final version of styles lesson template

tobyhodges:
Bringing the lesson template up-to-date in preparation for the planned infrastructure transition.

### DIFF
--- a/_includes/life_cycle.html
+++ b/_includes/life_cycle.html
@@ -28,7 +28,55 @@
 {% elsif site.life_cycle == "stable" %}
 
 {% comment %}
-We do not do anything special unless this is a Carpentries Lab lesson
+We do not do anything special for lessons in stable
+{% endcomment %}
+
+
+{% comment %}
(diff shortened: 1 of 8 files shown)


_includes/life_cycle.html
```diff
@@ -28,7 +28,55 @@
 {% elsif site.life_cycle == "stable" %}
 
 {% comment %}
-We do not do anything special unless this is a Carpentries Lab lesson
+We do not do anything special for lessons in stable
```
LilithElina, line 31: For lessons in stable... what?

LilithElina (approved):

_includes/life_cycle.html
tobyhodges: "stable" here refers to the last stage in [the lesson life cycle described in The Carpentries Curriculum Development Handbook](https://cdh.carpentries.org/the-lesson-life-cycle.html). This stage is used when the lesson is considered to be fully released i.e. out of beta testing.

PR #571, opened 2023-06-05
fixes a reference to an outdated error type

### LINKED ISSUE #549: error type raised by pandas has changed
In [_Data Types and Formats_, section _Missing Data Values - NaN_](https://datacarpentry.org/python-ecology-lesson/04-data-types-and-format.html#missing-data-values---nan), the error type mentioned as being raised has changed. The episode text [currently states](https://github.com/datacarpentry/python-ecology-lesson/blob/6e4b73165f998a7e2ada3853b19d3e876c04dda2/episodes/04-data-types-and-format.md?plain=1#L256):

> Notice that this throws a value error: `ValueError: Cannot convert NA to integer`.

but in recent versions of pandas the type of error thrown is much more specific `pandas.errors.IntCastingError: Cannot convert non-finite values (NA or inf) to integer`

The episode should be updated to reflect this.

Simon-Stone:
Fixes #549 by updating the reference to the error in the instructional text. To avoid confusion, the deprecated behavior (raising a `ValueError`) is also mentioned.

This is a contribution as part of my instructor checkout.

### DIFF
--- a/episodes/04-data-types-and-format.md
+++ b/episodes/04-data-types-and-format.md
@@ -132,7 +132,7 @@ surveys_df.dtypes
 
 which returns:
 
-```python
+```python 
 record_id            int64
 month                int64
 day                  int64
@@ -259,7 +259,7 @@ surveys_df['plot_id'].astype("float")
 2         2.0
 3         7.0
 4         3.0
-         ...
+         ... 
 35544    15.0
 35545    15.0
 35546    10.0
--- a/episodes/04-data-types-and-format.md
+++ b/episodes/04-data-types-and-format.md
@@ -68,12 +68,12 @@ numbers can not be used for mathematical operations**!
 pandas and base Python use slightly different names for data types. More on this
 is in the table below:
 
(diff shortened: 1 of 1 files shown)


btovar:
@Simon-Stone Thanks for your contribution!  Your commit include changes unrelated to the bug fix, would you know how to clean up these changes and only submit the relevant change? If not, that's something I can do for you.

Simon-Stone:
> Your commit include changes unrelated to the bug fix, would you know how to clean up these changes and only submit the relevant change? If not, that's something I can do for you.

It looks like those are some trailing whitespaces that my IDE cleaned up automatically. Do you want me to add those back in and push another commit?

btovar:
@Simon-Stone Yes, that sounds the easiest. I can then squash the pr into a single commit when merging.

Simon-Stone:
> @Simon-Stone Yes, that sounds the easiest. I can then squash the pr into a single commit when merging.

I hope this fixes it. Let me know if it needs any further changes!

btovar:
@Simon-Stone Looks good, thanks again!

LilithElina (approved):

PR #575, opened 2023-07-12
Fix environment link in setup page

### LINKED ISSUE #574: Environment link is broken
The link to download the environment file in the [Summary and Setup](https://datacarpentry.org/python-ecology-lesson/#alternative-installing-required-packages-with-environment-file) is not working.

The problem is in the line 196 of the file `learners/setup.md` that can be access via this [permalink](https://github.com/datacarpentry/python-ecology-lesson/blob/fbb0e211d0f795917af80144c3ffde5d1680639e/learners/setup.md?plain=1#L196C1-L196C74)

I believe the problem can be solved using the file path, something like:

```markdown
[environment.yml](episodes/files/environment.yml)
```

The complete link should be generated automatically when deploy the material.

patricia-ternes:
Fixes #574 by updating the path to the environment file.

_This is a contribution as part of my instructor checkout._

### DIFF
--- a/learners/setup.md
+++ b/learners/setup.md
@@ -193,7 +193,7 @@ conda install -c conda-forge plotnine
 ### *(Alternative)* Installing required packages with environment file
 
 Download the
-[environment.yml](https://episodes/files/environment.yml/environment.yml)
+[environment.yml](/files/environment.yml)
 file by right-clicking the link and selecting save as.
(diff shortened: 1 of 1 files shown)


btovar:
@patricia-ternes thanks!  It seems that the / was not needed. Let me investigate.

PR #584, opened 2024-01-30
Update join types in pandas merge function

### LINKED ISSUE #583: describe better joins
### How could the content be improved?

I think there is a significant opportunity to describe better joins here. Inner joins and left joins are described in detail. Right and full joins are defined in a smaller section. It would be helpful to describe all types of joins, including cross and self-join. I recommend adding:
First, cross join:
Cross-join generates the Cartesian product of the two data frames, producing all possible combinations of rows. The resulting product is a number of rows equal to the product of the number of rows in both data frames. While this join often results in large output, it can be helpful to see all the qualities of both tables, including each common and duplicate column.
Second, self-join. Self-join joins a data frame with itself. Self-join can be useful when you want to, for instance, compare records within the same dataset based on a given criteria.
Additionally, figures such as the ones shown here could be added: https://javarevisited.blogspot.com/2013/05/difference-between-left-and-right-outer-join-sql-mysql.html#axzz8PwmFS4FN

Submitted on behalf of trainee for checkout. 

### Which part of the content does your suggestion apply to?

https://datacarpentry.org/python-ecology-lesson/05-merging-data.html

quist00:
Closes #583

Expanded description of cross join with suggestions and added self-joins with a resource for users to explore that further on their own time if they need that approach.  Joins is already a relatively heavy subject for learners to grapple with so opted not to have full pedagogical sections covering these additional join types in detail.

### DIFF
--- a/episodes/05-merging-data.md
+++ b/episodes/05-merging-data.md
@@ -434,16 +434,20 @@ case, `PF`) does not occur in `species_sub`.
 
 ### Other join types
 
-The pandas `merge` function supports two other join types:
+The pandas `merge` function supports other join types:
 
 - Right (outer) join: Invoked by passing `how='right'` as an argument. Similar
   to a left join, except *all* rows from the `right` DataFrame are kept, while
   rows from the `left` DataFrame without matching join key(s) values are
   discarded.
 - Full (outer) join: Invoked by passing `how='outer'` as an argument. This join
-  type returns the all pairwise combinations of rows from both DataFrames; i.e.,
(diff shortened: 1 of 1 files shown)


quist00:
@sheraaronhurt assuming I understand this comment correctly "Submitted on behalf of trainee for checkout," is there github username or other identifier for that learner we can include in the comment stream for reference purposes?

sheraaronhurt:
@quist00 Here is the information requested: [rmr969](https://github.com/rmr969)

PR #631, opened 2025-08-18
Visualization matplotlib

josenino95:
Draft on "Data visualization with Pandas and Matplotlib" episode. Now that I'm creating this PR, I think that even the episode name can be improved. On the same lines as Toby's PR, here are some notes:
- It tries to mirror the [Data visualization with ggplot2](https://datacarpentry.github.io/R-ecology-lesson/instructor/visualizing-ggplot.html) episode from the DC Ecology R lesson.
- While I was writing it I realized exactly mirroring the R lesson would not be possible, as ggplot code can be very different (and simpler) than matplotlib in some cases. Specially for overlaying making jitter plots. Therefore I turned to keep code simple and show some interesting stuff that can be done with matplotlib, like subplots and sharing axis between subplots.
- At the end I felt the need to mention other popular plotting libraries. But if you feel that is too much content, we can delete it. Or improve the content.
- I also took a swing at estimating the time of the lesson with exercises.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -69,6 +69,7 @@ episodes:
 - 07-visualization-ggplot-python.md
 - 08-putting-it-all-together.md
 - 09-working-with-sql.md
+- visualizing-matplotlib.md
 
 # Information for Learners
 learners: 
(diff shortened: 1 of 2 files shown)


josenino95:
Sorry, this PR was meant to be for the incubator fork we are working on for the revamp of the lesson.

